PR #64, opened 2023-05-02
PuterBot-46: Add Publishing using magic-wormhole

### LINKED ISSUE #46: Implement Publishing
We would like to make it easy to publish `puterbot.db`:

```
python -m puterbot.publish  "<task_description>" [<puterbot.db>]
```

How can we store this in a decentralized way? e.g.:

- [IPFS](https://ipfs.tech/)
- [Sia](https://sia.tech/)
- [FileCoin](https://filecoin.io/)
- [Arweave](https://www.arweave.org/)
- others?

Mustaballer:
## Overview

- Resolves #46 
- Using https://github.com/magic-wormhole/magic-wormhole to send and receive recordings

## Unit Tests Created

- will do after review & iteration

## Steps to QA

- After creating a recording, run these following commands in yout venv environment
- For sending a recording: `python -m puterbot.share send --recording_id=1 --output_folder=output`
- For receiving a recording: `python -m puterbot.share receive --output_folder=output`


abrichr:
Let's replace ` --output_folder=output` with `config.RECORDING_DIR_PATH`

Mustaballer:
I was doing some research on ways to store images in a db(specifically the image differences), and I was wondering what you think of using AWS S3. We can store the image data in an external storage service like S3 and store only the reference (e.g., S3 object key) in the database. This allows us to efficiently retrieve and serve the images in the frontend by directly accessing them from the CDN or S3 URL based on the reference stored in the database. 

```markdown
### ChatGPT code

```python
import boto3
from botocore.exceptions import NoCredentialsError

class Screenshot(Base):
    __tablename__ = "screenshot"

    id = sa.Column(sa.Integer, primary_key=True)
    recording_timestamp = sa.Column(sa.Integer)
    timestamp = sa.Column(sa.Integer)
    image_key = sa.Column(sa.String)  # Store the reference to the image in the S3 bucket

    prev_timestamp = sa.Column(sa.Integer)

    prev = None
    _image = None
    _diff = None
    _diff_mask = None

    def save_to_s3(self):
        # Connect to the S3 bucket
        s3 = boto3.client("s3")

        # Generate a unique key for the image in the S3 bucket
        image_key = f"screenshot_{self.id}.png"

        # Save the image to the S3 bucket
        try:
            s3.put_object(
                Body=self.image.tobytes(),
                Bucket="your-s3-bucket-name",
                Key=image_key
            )
        except NoCredentialsError:
            print("Error: AWS credentials not found. Unable to save to S3.")
            # Handle the exception according to your error handling strategy

        # Update the image_key attribute with the reference to the image in the S3 bucket
        self.image_key = image_key

    @property
    def image(self):
        if not self._image:
            # Fetch the image from the S3 bucket using self.image_key
            # Implement the logic to retrieve the image from S3 and return it as an Image object
            pass

        return self._image

    # ... other properties and methods ...
```

### Benefits to this approach(From ChatGPT)

1. **Scalability and Performance**: Storing images directly in the database can increase its size and impact performance. By storing images in an S3 bucket or CDN, you can offload the storage and delivery of images to a dedicated service that is optimized for handling large-scale data and providing high-performance image serving.

2. **Cost-Effectiveness**: External storage services like Amazon S3 offer cost-effective storage solutions, allowing you to scale and manage your image storage more efficiently. It eliminates the need for storing large binary data directly in the database, which can lead to increased storage costs.

3. **Flexibility and Accessibility**: By storing the reference to the image (e.g., S3 object key) in the database, you can easily retrieve and serve the images in your frontend applications by directly accessing them from the CDN or S3 URL based on the stored reference. This provides flexibility in integrating with different systems and simplifies the image retrieval process.

4. **Ease of Maintenance**: Storing images in a dedicated storage service decouples the image storage from the database, resulting in a cleaner and more maintainable database structure. It allows you to focus on the core database operations without worrying about managing binary image data.

Mustaballer:
@abrichr Currently, I am successful in sending a compressed zipped file using magic-wormhole!

This image is for sending the zipped file, it includes zipping the recording screenshots --> sending the zipped file in a wormhole --> completes when the zipped file is receieved.
image

Question/Issue:
- Right now I am sending every screenshot associated with a recording as a compressed zip file. I know the plan is to store the image diffs in the db, is that also what we want to share across computers as opposed to a db file?
- I currently use the highest compression for these screenshots, but when I extract these images after receiving the zip file, I am unable to view the png, I get file type is unsupported. Without compressing the zip file I am able to view the pngs after extracting but it ends up being minimum a GB in storage.

Mustaballer:
@abrichr Currently, I am using magic-wormhole to transfer the entire database file. I attempted to create a new database file and add the recording to it, but encountered errors in the process. It appears that there is a workaround available by marking the recording as transient and adding it to the new database. However, this approach removes the recording from the old database, which is not the desired outcome. As a temporary solution, we can continue sending the entire database file. What are your thoughts on this?

abrichr:
@Mustaballer can you please clarify what it means to make a recording as transient?

I think what we want is this:

1. Export to sql:
```
def export_sql(recording_id):
    # Export data for recordings e.g. via SELECT or similar

```
2. Create new database
```
def create_db(recording_id):
    fname_parts = [
        config.DB_FNAME,
        recording_id,
        datetime.now().strftime(config.DT_FMT),
    ]
    db_fname = "-".join(fname_parts)
    config.set_db_fname(db_fname)

config.py:
```
def set_db_fname(db_fname):
    global DB_FNAME
    DB_FNAME = db_fname
    set_db_fpath()

def set_db_fpath():
    global DB_FPATH
    DB_FPATH = ROOT_DIRPATH / DB_FNAME

def set_db_url():
    global DB_URL
    DB_URL = f"sqlite:///{DB_FPATH}"
    logger.info(f"{DB_URL=}")

set_db_url()
```

set_db_host()
set_db_url()

**Comment by abrichr** [comment]
Hi @Mustaballer what is the status of this?

**Comment by Mustaballer** [comment]
I will create a new pull request to address the changes and remove all the formatting modifications. Additionally, I will use a different style, such as Google Python, instead of reStructuredText after confirming with you.

**Comment by Mustaballer** [comment]
@abrichr I resolved the issue with creating a new db file, the issue I currently face is importing the recording to the new db file. This code executes a SQL query on the db, however it doesn't actually import the data
```python
def run_sql(sql):
    with db.engine.connect() as con:
        result = con.execute(sql)
        for row in result:
            logger.info(f"{row=}")
```

Mustaballer:
Will close this PR, and open a new one without all the formatting changes

Issue #24, opened 2023-04-19
Expand on DemoReplayStrategy


abrichr:
How can we modify the [DemoReplayStrategy](https://github.com/MLDSAI/puterbot/blob/main/puterbot/strategies/demo.py) that uses the available primitives (e.g. LLMs, OCR, SVG, etc) in a way that makes sense?

dianzrong:
Hi, 
I would love to work on this issue, however, I'm confused as to what this issue is asking for. Could I get some further clarification? 
Thanks!

abrichr:
HI @dianzrong, thank you for your interest in this issue!

From [demo.py](https://github.com/MLDSAI/puterbot/blob/main/puterbot/strategies/demo.py):

```
    def get_next_input_event(
        self,
        screenshot: Screenshot,
    ):
        ascii_text = self.get_ascii_text(screenshot)
        #logger.info(f"ascii_text=\n{ascii_text}")

        ocr_text = self.get_ocr_text(screenshot)
        #logger.info(f"ocr_text=\n{ocr_text}")

        event_strs = [
            f"<{event}>"
            for event in self.recording.input_events
        ]
        history_strs = [
            f"<{completion}>"
            for completion in self.result_history
        ]
        prompt = " ".join(event_strs + history_strs)
        N = max(0, len(prompt) - MAX_INPUT_SIZE)
        prompt = prompt[N:]
        logger.info(f"{prompt=}")
        max_tokens = 10
        completion = self.get_completion(prompt, max_tokens)
        logger.info(f"{completion=}")

        # only take the first <...>
        result = completion.split(">")[0].strip(" <>")
        logger.info(f"{result=}")
        self.result_history.append(result)

        # TODO: parse result into InputEvent(s)

        return None
```

This code is converting the screenshot into text using ASCII and OCR, but the values of `ascii_text` and `ocr_text` are not used when generating a completion. This means that the language model is generating completions solely based on the `InputEvent`s in the recording, ignoring the screenshots altogether.

Imagine if a person were asked to accomplish a task by only looking at input events. Maybe they could memorize them, but if anything changed on the screen they wouldn't know. The same is true for the model.

This task involves thinking about how to combine the `ocr_text` and/or `ascii_text` with the `InputEvent`s to generate the prompt.

Does that clarify things?

dianzrong:
@abrichr, yes that does. Thank you so much!!

PR #89, opened 2023-05-06
Detect diffs via ASCII/OCR

AvidEslami:
Progress towards: #24 
Summary:
Use OCR and ASCII data to determine when an input event has occurred,
         example: task is to open task manager (press cmd, then search for task, then press enter)
                        If cmd window is accidentally closed before search is completed, then the task will repeat and open cmd again. 
                        Once the open cmd window is active, the search will be completed and task manager will be opened.
Implementation:
 - Compares OCR and ASCII data between replay iterations to determine when visual changes occur.

Assumptions:
 - OCR and ASCII data can be used to determine when an input event has occurred (Inputs have visible impact on screen).
 - Screen is not changing due to background processes (e.g. a video is playing in the background).

 TODO:
 - Generate InputEvents from completion results as opposed to previous recordings.
 - Gather OCR and ASCII data of WindowEvents as opposed to entire screen (get_active_window, issue 84).


abrichr:
Thank @AvidEslami ! 

> Compares OCR and ASCII data between replay iterations to determine when visual changes occur.

Can you please clarify why we want to look at the diff of the OCR/ASCII and not just the screenshot image itself?

AvidEslami:
> Can you please clarify why we want to look at the diff of the OCR/ASCII and not just the screenshot image itself?

Good point @abrichr ,

Since the images are stored using PIL (python imaging library) we can compare them directly, though I found advantages and disadvantages with this implementation.

Advantages of comparing images directly:
 - Captures all visual changes including fine details
 - **If** OCR and ASCII data are **not** needed elsewhere within the application, comparing images directly is more efficient

Advantages of comparing OCR and ASCII data:
 - Less sensitive to minor differences in pixel values (Visual artifacts)
 - **If** OCR and ASCII data are available, less comparisons are required compared to high resolution images (comparing individual pixels)

**Summary:** Using OCR and ASCII data allows us to focus on textual and structural differences while reducing the impact of minor rendering details. For the purposes of OpenAdapt, OCR and ASCII data could additionally be used elsewhere within the task such as when generating completions giving us a potential need to compute them regardless. Therefore using OCR and ASCII data requires less comparisons with reference to using Image data directly, example: 1920x1080 screen resolution (2,073,600 pixel comparisons), vs. 120x30 ASCII + variable OCR size (3600 character comparsions + OCR)

abrichr:
> Therefore using OCR and ASCII data requires less comparisons with reference to using Image data directly, example: 1920x1080 screen resolution (2,073,600 pixel comparisons), vs. 120x30 ASCII + variable OCR size (3600 character comparsions + OCR)

Good point. How does this comparison change if we consider the additional computation required by OCR/ASCII?

AvidEslami:
> Good point. How does this comparison change if we consider the additional computation required by OCR/ASCII?

Here's some Data I was able to gather:

**Case: 1920x1080 screen resolution, 2675 characters on screen**
Average time to compute and compare ascii and OCR: 18.382150602340698
Average time to compare ascii and OCR: 0.002645111083984375
Average time to compare images: 0.024423050880432128
Average time to compute ascii alone: 0.2591431140899658
Average time to compute OCR alone: 18.038353204727173

On my local device, OCR is taking significantly longer than ASCII

abrichr:
It looks like comparing images directly is faster than doing it via ASCII or OCR. Do you agree? If so, should we close this?

Issue #47, opened 2023-05-01
Implement Scrubbing


abrichr:
We would like to make it easy to scrub private information (e.g. PII/PHI):

```
python -m openadapt.scrub
```

KrishPatel13:
Dear @abrichr,

I would like to work on implementing this issue. Just to confirm, my understanding is that we need to scrub private information from the input text such as:
- Date of Births
- Phone Numbers
- Bank Card Numbers
- Email Addresses

Please let me know if there are any additional requirements or specifications that I should consider while implementing the solution.

abrichr:
Excellent, thank you Krish! These are the top Google results for [pii python], this looks promising, please start here:

https://pypi.org/project/piicatcher/
https://pypi.org/project/pii-extract/

Here is a Phind.com search that may be insructive:

https://www.phind.com/search?q=how%20to%20implement%20a%20PII%2FPHI%20scrubber%20in%20python%20for%20text%20and%20images%3F%20we%20already%20have%20OCR%20available&c=&source=searchbox

Please provide a few options and tradeoffs for each.

💪 let's do this 💪

abrichr:
Phind.com recommends https://github.com/microsoft/presidio. After a glance it seems the top candidate (please let me know if you identify any other candidates worth comparing).

In your analysis please include a short description of the available APIs, e.g. function signatures, data sources/formats.

Thank you!!

KrishPatel13:
Here's my analysis of the three options you've mentioned for implementing a PII/PHI scrubber in Python, along with their tradeoffs:

### **piicatcher:**

- piicatcher is a Python package for detecting personally identifiable information (PII) in text using regular expressions.
- It currently supports the detection of email addresses, credit card numbers, US phone numbers, US Social Security numbers, and IP addresses.
- The package is easy to install and use, and it provides a simple API for detecting PII in text.
- Tradeoff: it only supports a limited number of PII types, and it relies solely on regular expressions for detection, which can result in false positives or missed detections.
- Sample Usage API:
    • Function Signature: 
        `def catch(text: str, include_pattern_names: bool = False) -> List[str]:`
    • The function takes in a string of text as input and returns a list of PII entities detected in the text.


### **pii-extract:**

- pii-extract is a Python package for extracting PII from text using machine learning models.
- It supports the extraction of various types of PII, including email addresses, phone numbers, credit card numbers, and dates of birth.
- The package is easy to install and use, and it provides a simple API for extracting PII from text.
[...]

PR #116, opened 2023-05-08
Keyboard Input Scrubbing

micol-altomare:
This pull request is a proof-of-concept of a #47 scrubbing implementation for keyboard input only (for now). It considers keyboard events in batches and uses Regex to detect and filter out sensitive information (for now, phone numbers and the number "1"; this can easily be scaled to detect more types of sensitive information). See my [readme.md](https://github.com/micol-altomare/OpenAdapt/blob/scrubbing/readme.md) for more information.

### DIFF
--- a/puterbot/record.py
+++ b/puterbot/record.py
@@ -141,8 +141,9 @@ def process_events(
                 prev_saved_window_timestamp = prev_window_event.timestamp
         else:
             raise Exception(f"unhandled {event.type=}")
-        prev_event = event    
+        prev_event = event
     logger.info("done")
+
 
 def write_input_event(
     recording_timestamp: float,
--- a/puterbot/record.py
+++ b/puterbot/record.py
@@ -141,9 +141,8 @@ def process_events(
                 prev_saved_window_timestamp = prev_window_event.timestamp
         else:
             raise Exception(f"unhandled {event.type=}")
-        prev_event = event
+        prev_event = event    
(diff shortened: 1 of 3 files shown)


abrichr:
Thank you @micol-altomare !

Do you know why the diff shows the whole record.py as being changed?

Also, if you click through to the issue there is some discussion there around alternative approaches using standard libraries. It's probably best to sync with others who are working on it so as not to duplicate effort.

Please fix the issue causing in the enormous diff, then I can take a look and we can go from there.

Thank you!

0dm:
@micol-altomare  My hypothesis for the large diffs is that it's something to do with Windows vs Mac/Linux, more specifically the line termination sequences being written as CR LF vs LF. I've never encountered this before but I found a similar problem here: https://stackoverflow.com/questions/19593909/git-diff-sees-whole-file-as-changed-when-its-not

micol-altomare:
Hello @abrichr and @0dm, thank you for the guidance and suggestion. The diff should be fixed now!

abrichr:
Thank you @micol-altomare ! 

It looks like @KrishPatel13 was already assigned to this issue, and this approach is already deprecated as per the comments on the issue, so unfortunately we won't be able to merge this.

Can you please pick 2-3 other issues that are not already assigned to anyone, and comment on them with your suggested approach?

Thank you!

Issue #85, opened 2023-05-05
Save perf_q to database


abrichr:
This task involves creating a model and saving events from `perf_q` in `record.py`:

https://github.com/MLDSAI/puterbot/blob/main/puterbot/record.py#LL591C10-L591C10

abrichr:
@apgorton thank you for finding the source of https://github.com/MLDSAI/OpenAdapt/issues/93 ! I believe this is the solution. Let's treat performance events just like Screenshots/InputEvents/WindowEvents and save them to a new table. That way there won't be any special queue left over when everything is shutting down and that problem goes away.

What do you think?

apgorton:
I think this approach would address https://github.com/MLDSAI/OpenAdapt/issues/93 and is a good way to do that while implementing this one. I'll make the changes.

PR #122, opened 2023-05-08
Save performance data to db

### LINKED ISSUE #93: Record hanging on windows instead of ending
Windows 10

Sometimes on windows record.py will hang after Ctrl-C with python processes left running.

apgorton:
Save the perf_q event performance data to db for #85

Fixes #93


puterbot/crud.py
```diff
@@ -68,6 +68,46 @@ def insert_window_event(recording_timestamp, event_timestamp, event_data):
     }
     _insert(event_data, WindowEvent, window_events)
 
+def insert_perf_stat(recording_timestamp, event_type, start_time, end_time, buffer=None):
```
abrichr, line 71: Why not re-use `_insert` here, like the other events?

puterbot/record.py
```diff
@@ -431,10 +432,35 @@ def read_window_events(
         prev_geometry = geometry
 
 
-def plot_performance(
-    recording_timestamp: float,
+def performance_stats_writer (
```
abrichr, line 435: Why not re-use `write_events` here?

puterbot/record.py
```diff
@@ -668,6 +693,17 @@ def record(
     )
     window_event_writer.start()
 
+    terminate_perf_event = multiprocessing.Event()
```
abrichr, line 696: Why not re-use `terminate_event` here?

abrichr (changes requested):
Thank you for putting this together @apgorton ! Just a few small comments, please let me know your thoughts!

puterbot/record.py
apgorton: We need a separate multiprocessing.Event() so that the event writers finish putting to perf_q before the performance stat writer terminates. Otherwise we risk losing data and risk a deadlock like #93.

puterbot/crud.py
apgorton: Changes made with latest commit.

puterbot/models.py
```diff
@@ -222,3 +222,11 @@ class WindowEvent(Base):
     top = sa.Column(sa.Integer)
     width = sa.Column(sa.Integer)
     height = sa.Column(sa.Integer)
+
```
abrichr, line 225: Please add two newlines between classes

abrichr:
Thanks @apgorton ! Can you please set PLOT_PERFORMANCE = True and paste the result here to confirm it's working?

puterbot/models.py
apgorton: Why is that?

apgorton: I see the file is like that already, done in latest commit.

apgorton:
> Thanks @apgorton ! Can you please set PLOT_PERFORMANCE = True and paste the result here to confirm it's working?

performance-1683654095 646531

puterbot/crud.py
abrichr: Thank you @apgorton ! In the future please include the commit hash in your comment so that the reader can easily click through 🙏

puterbot/models.py
abrichr: Thanks @apgorton ! See https://stackoverflow.com/a/47448739/95989 . From [PEP8](https://peps.python.org/pep-0008/#imports):

> Surround top-level function and class definitions with two blank lines.

puterbot/record.py
```diff
@@ -16,7 +16,6 @@
 import sys
 import threading
 import time
-import zlib
```
abrichr, line 19: Ninja

abrichr:
Thank you @apgorton ! In the future please don't forget to include your alembic scripts (see https://github.com/MLDSAI/OpenAdapt#generate-migration-after-editing-a-model)

before: (08dac6)
image

after: (4aca87d)
image

We would like to see a before and after, but this "before" is on the latest main, which has a whole lot of other changes since this branch was created (working to improve turnaround time).

I think before we merge this we want to understand what is the performance impact 😅

abrichr:
Edit: real before and after

Before:
image

After: 
image

Looks like performance has improved! Thank you @apgorton ! 🙏 💯 🚀

PR #125, opened 2023-05-09
#127 LLMReplayStrategyMixin class: Fixed prompt truncation and type contract

### LINKED ISSUE #127: Incorrect Truncation of Prompt in LLMReplayStrategyMixin class.
As the title suggests, the prompt string is being truncated incorrectly inside the ```get_completion``` method in the ```llm_mixin.py``` file. The first [0, max_input_size-1] characters must be fetched but instead the characters in the range [max_input_size, len(prompt)-1] are being fetched.

FFFiend:
Resolved #127 
Simple fix, truncated prompt correctly, and set type contract of max_input_size inside the ```LLMReplayStrategyMixin``` class's initializer params as ```int```

### DIFF
--- a/puterbot/strategies/llm_mixin.py
+++ b/puterbot/strategies/llm_mixin.py
@@ -25,7 +25,7 @@ class LLMReplayStrategyMixin(BaseReplayStrategy):
         self,
         recording: Recording,
         model_name: str = MODEL_NAME,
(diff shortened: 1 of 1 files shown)


abrichr:
Thank you @FFFiend !

I believe this implementation is not quite right; see https://github.com/MLDSAI/OpenAdapt/issues/127#issuecomment-1540112308

PR #147, opened 2023-05-14
Added OpenAI support to LLM Mixin

metalfilter:
The system defaults to gpt-3.5-turbo on OpenAI.
The original Hugging Face support is still available. The LLM mixin now takes an additional platform variable. The use of OpenAI as a platform requires:
 - The OPENAI_SYSTEM_MESSAGE be set in config.py
 - The OpenAI API key be included in keys.json in the root folder

Note that the keys.json file should have the following format: '{ "openai": "[key]" }'

### DIFF
--- a/puterbot/config.py
+++ b/puterbot/config.py
@@ -1,3 +1,4 @@
+import json
 import pathlib
 
 ROOT_DIRPATH = pathlib.Path(__file__).parent.parent.resolve()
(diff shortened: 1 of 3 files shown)


abrichr:
Thank you @metalfilter ! This functionality is contained in https://github.com/MLDSAI/OpenAdapt/pull/151 .

In the future please check to see if there is a relevant issue and/or PR so as not to duplicate effort 🙏

PR #158, opened 2023-05-17
modal.com

### LINKED ISSUE #113: Explore modal.com
How can we take advantage of modal.com? They've given us free API credits for the Tribe AI LLM hackathon.
image

FFFiend:
Resolves #113 

Upon running ```python modal_completions.py``` users can see the actual completions of the input prompt (can be modified in file, however mixins aren't supposed to be run directly and this is an enhancement for ```llm_mixin.py``` to potentially support generation from larger LLMs by having the completions done on the cloud instead of your local machine, and modal_completions exists purely for demonstration purposes)

Prerequisites: Please follow the instructions over at [modal](https://modal.com/docs/guide) and generate your api token, before running the aforementioned command to view the completion. 

Note that the completion is done only on gpt2 for now since that's the quickest proof of concept, however larger models can be loaded by modifying the gpu type, memory and count. Also, make sure to increase the timeout since the process may take longer (initializing the image as well as the model ON the image)

Additionally, ```modal_plugin.py``` is now fairly versatile since we now have an entry point, of sorts for use of multiple generative models. Further enhancements may look something like: adding code inside the ```complete``` method for initializing and generating completions on common/famous LLMs such as T5, Bert, etc.

**Note also that using GPUs costs money, however every Modal user starts off with a 30 USD balance that they can use to take these GPUs for a test drive. Also note that these GPUs are billed per hour of use.***

Do let me know if you run into any errors.


FFFiend:
Note that if the current GPU is busy, there is a queue time (unfortunate downside) and also, loading larger models can be rather slow which is why the process timeout needs to be increased. Luckily, modal timeouts go to a max of 24 hours. Speed is a tradeoff here but at least we can run models and generate completions from the ones with high minimum memory requirements using Modal GPUs.

abrichr:
Excellent work @FFFiend ! What's next?

FFFiend:
Well I wanna try to finish off Gisting but I haven't been able to tinker around too much this week. 

The problem with Gisting is that the files are dense and interconnected and so whipping up a Modal instance was not so easy, but I think if I spend a while with the file system I can host the entire thing on the cloud with some smart class instances and then we can close that issue for good.

PR #174, opened 2023-05-26
Layout Extraction and Modal.com

FFFiend:
Addresses #187 , coupled with Modal.com files

The file added as ```test.png``` is a screenshot taken of a publicly available, contract template from Google images.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1766,6 +1766,120 @@ files = [
 ]
 
 [[package]]
+name = "levenshtein"
+version = "0.21.1"
+description = "Python extension for computing string edit distances and similarities."
+optional = false
+python-versions = ">=3.6"
+files = [
+    {file = "Levenshtein-0.21.1-cp310-cp310-macosx_10_9_universal2.whl", hash = "sha256:59e5054c9dea821840af4623a4059c8f0ae56548a5eae8b9c7aaa0b3f1e33340"},
+    {file = "Levenshtein-0.21.1-cp310-cp310-macosx_10_9_x86_64.whl", hash = "sha256:11694c6f7119d68cc199ff3b1407560c0efb0cc49f288169f28b2e032ee03cda"},
+    {file = "Levenshtein-0.21.1-cp310-cp310-macosx_11_0_arm64.whl", hash = "sha256:f5f7ce639bea0f5e95a1f71963624b85521a39928a2a1bb0e66f6180facf5969"},
+    {file = "Levenshtein-0.21.1-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:39e8a1866325b6d54de4e7d1bffffaf4b4c8cbf0988f47f0f2e929edfbeb870d"},
+    {file = "Levenshtein-0.21.1-cp310-cp310-manylinux_2_17_ppc64le.manylinux2014_ppc64le.whl", hash = "sha256:ed73d619e203aad54e2e6119a2b58b7568a36bd50a547817d13618ea0acf4412"},
+    {file = "Levenshtein-0.21.1-cp310-cp310-manylinux_2_17_s390x.manylinux2014_s390x.whl", hash = "sha256:50fbe01be99554f644657c32a9e3085369d23e8ccc540d855c683947d3b48b67"},
+    {file = "Levenshtein-0.21.1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl", hash = "sha256:675ba3afaa9e8ec393eb1eeee651697036e8391be54e6c28eae4bfdff4d5e64e"},
+    {file = "Levenshtein-0.21.1-cp310-cp310-manylinux_2_5_i686.manylinux1_i686.manylinux_2_17_i686.manylinux2014_i686.whl", hash = "sha256:c89a5ac319a80c131ca8d499ae0f7a91d4dd1dc3b2e9d8b095e991597b79c8f9"},
+    {file = "Levenshtein-0.21.1-cp310-cp310-musllinux_1_1_aarch64.whl", hash = "sha256:f9e3a5f4386c8f1811153f309a0ba3dc47d17e81a6dd29aa22d3e10212a2fd73"},
+    {file = "Levenshtein-0.21.1-cp310-cp310-musllinux_1_1_i686.whl", hash = "sha256:ea042ba262ea2a95d93c4d2d5879df956cf6c85ce22c037e3f0d4491182f10c5"},
+    {file = "Levenshtein-0.21.1-cp310-cp310-musllinux_1_1_ppc64le.whl", hash = "sha256:622bc670b906c4bf219755625e9fa704ff07c561a90f1aa35f3f2d8ecd3ec088"},
+    {file = "Levenshtein-0.21.1-cp310-cp310-musllinux_1_1_s390x.whl", hash = "sha256:f0e51ff6d5665884b0e39b4ae0ef4e2d2d0174147147db7a870ddc4123882212"},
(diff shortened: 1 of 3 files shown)


FFFiend:
https://huggingface.co/impira/layoutlm-document-qa 

This model is fine tuned for document question answering, and uses the same LayoutLM model to do so. It solves multiple problems such as conversion, data extraction etc but I'm testing it and it seems like its tuned to only give one word responses at max? See below:
```
nlp = pipeline("document-question-answering",
model = "impira/layoutlm-document-qa")
print(nlp("contract.png","Could you extract all key information from this document?"))

#output below 
[{'score': 0.0048963166773319244, 'answer': 'bills', 'start': 14, 'end': 14}]
```

FFFiend:
As for parsing the output yourself, seems like we need to do training and fine tuning to fit the freight use case, after looking at a few examples online people use predefined labels and then use LayoutLM to extract predictions of those labels within the document, rather than extracting actual text from the output.

FFFiend:
Here is the output you get after decoding the input ids for the Freight Contract document.
```
[CLS] hazardous material manifest truck : route truck 001 ( rt001 ) date : 9 / 25 / 2015 7 : 50 : 02 am manifest # : rt00120150925075002 departing location : driver : proper shipping name hazard pg rq quantity returned package special product code total weight class quantity permit ( lbs ) un proper shipping name hazard pg rq quantity returned package special product code total weight number class quantity pennit ( lbs ) sy ee, | es es ss | 5 a es [ 0 | _ | eee [ so es se ) ss ss es 2 | [SEP]
```

FFFiend:
Screen Shot 2023-06-02 at 8 33 26 PM 

An image of how bounding boxes are shown on a load manifest document. It's not grabbing everything probably because the resolution of the file isn't small enough to be tokenized into 512 tokens. See below for a better result on a rate confirmation document.


Screen Shot 2023-06-02 at 8 35 05 PM

FFFiend:
As for classification, this has to be done manually. Labels and colors must be initialized and then a correspondence has to be made between the boxes, input_ids and the labels. For example: {TABLE:5, CONTACT_INFO:3. ..}  where we have a frequency dict of each type of structure. Will look into this now

FFFiend:
Screen Shot 2023-06-03 at 4 01 13 AM

Model mislabels mostly everything, as it encodes all the classified texts as 0's and 1's (which map to 'TABLE' and 'CONTACT_INFO' in my label_dict. Need to train on a freight dataset to get optimal results.

openadapt/strategies/mixins/freight_documents_test/Screenshot Dataset.v1i.coco/README.dataset.txt
```diff
@@ -0,0 +1,6 @@
+# Screenshot Dataset > 2023-06-03 1:20pm
```
abrichr, line 1: Can you please move this (and other similar files) into the `resources` directory?

FFFiend:
Currently experiencing an issue in regards to bounding boxes and words after processing data. Will prioritize today.

FFFiend:
> @FFFiend can you please prioritize merging this?
> 
> Also why are the aspect ratios off on the images (leading to them looking "squished")?

I think they were stored in the dataset on roboflow this way.

FFFiend:
Okay, so after coming back to this issue I realized that the above picture with just 1 type of label being classified was because I wasn't inferring properly. Here is a better result on an Invoice bill (in terms of classification):
Screen Shot 2023-06-17 at 4 24 59 AM

Results are not great... but it's a slight improvement from last time.

FFFiend:
Question I asked GPT-4 on LayoutLMv2's off the shelf labels. 
```
The LayoutLMv2 model is equipped with off-the-shelf labels that are based on the FUNSD dataset, which can be generalized for form-like tasks, such as information extraction from documents, receipts, etc. The dataset has the following 4 main labels:

1. "S": for the "selected" elements of forms, representing filled-in information.
2. "B": for "beginning", representing the starting token or element of a labeled field in a form.
3. "I": for "inside", representing the intermediate tokens or elements within a labeled field in a form.
4. "O": for "outside", representing tokens or elements that do not belong to any labeled field in a form.

These labels are a variant of the standard BILOU (Beginning, Inside, Last, Outside, Unit-length) tagging scheme – typically required for labeling document entities using the LayoutLMv2 model. Note that LayoutLMv2 adapts to a specific task when fine-tuned on annotated data related to that task.
```

Note that for GUI, this model is unsuitable because it focuses specifically on text classification and extraction. I think going more towards Object Detection is what we want. So current action items would be:
```
1) Fine tune LayoutLMv2 on label format and form dataset (lots available on paperswithcode) similar to what we have in the commit currently.

2) YOLO/ CNN for Object Detection including training + inference on labelled GUI data.
```
The above only takes care of labels, and not **tables**. There's models trained solely for table classification such as https://huggingface.co/microsoft/table-transformer-detection. This can be simply incorporated by passing the image once into LayoutLMv2 and then into a table transformer to extract both.

FFFiend:
Here's a couple HuggingFace spaces with different detection/extraction models:
https://huggingface.co/spaces/ShilongLiu/Grounding_DINO_demo

https://huggingface.co/spaces/nielsr/LayoutLMv2-FUNSD

FFFiend:
I think this might be what we need for GUI object detection..seems easier to setup than gisting :D but their demo isn't live 

https://github.com/chenjshnn/Object-Detection-for-Graphical-User-Interface

abrichr:
> I think this might be what we need for GUI object detection..seems easier to setup than gisting :D but their demo isn't live

I explored this about a year ago and didn't find much of value, author did not respond to emails. If you have any specific suggestions about how we can use it please clarify 🙏

abrichr:
> https://huggingface.co/spaces/ShilongLiu/Grounding_DINO_demo

Great find! This looks very promising. Added in https://github.com/MLDSAI/OpenAdapt/issues/288

abrichr:
> https://github.com/MLDSAI/OpenAdapt/issues/288

Can you please clarify how this is different from what you've already implemented?

FFFiend:
> > https://huggingface.co/spaces/nielsr/LayoutLMv2-FUNSD
> 
> Can you please clarify how this is different from what you've already implemented?

No difference, just a convenient demo hosted on the web to try out instead of having to run the script each time. Sorry i should've elaborated in the original comment.



> > I think this might be what we need for GUI object detection..seems easier to setup than gisting :D but their demo isn't live
> 
> I explored this about a year ago and didn't find much of value, author did not respond to emails. If you have any specific suggestions about how we can use it please clarify 🙏

There's a repo linked in the Readme that contains the model along with a script to test it but I think it needs some refactoring and package updates.

FFFiend:
Sure thing: Current action items are
```
1) Test and display output from Document Question-Answering Pipeline on the LayoutLMv3 model.
2) Train raw extraction on the LayoutLMv3 model.
```

FFFiend:
https://github.com/huggingface/diffusers/issues/3828 Currently facing some issues with using pipelines on LayoutLMv3.

FFFiend:
In the meantime, would you like to have a mixin that takes in an image along with a question (for document classification) using the original layout-lm model that seemed to be working well in the first reply to this PR?

abrichr:
@FFFiend can you please follow up on the issue you posted to huggingface/diffusers?

Also in the future please include the full exception trace, not just the error message.

> In the meantime, would you like to have a mixin that takes in an image along with a question (for document classification) using the original layout-lm model that seemed to be working well in the first reply to this PR?

Let's try to get v3 working 🙏

abrichr:
Closed in favor of https://github.com/OpenAdaptAI/OpenAdapt/pull/341

Issue #181, opened 2023-05-27
Investigate pyment


abrichr:
https://github.com/dadadel/pyment

PR #184, opened 2023-05-27
Investigated Pyment #181

0dm:
#181 
Ran `pyment -w openadapt` (including nested directories)

Please see the diff, it has added quite a bit, along with some empty docstrings for functions that have no params or return vals.

Are there any particular styles we want?

The current options are:
["javadoc", "reST", "numpydoc", "google"] (default "reST")

### DIFF
--- a/openadapt/cache.py
+++ b/openadapt/cache.py
@@ -9,18 +9,42 @@ from openadapt import config
 
 
 def default(val, default):
+    """
+
+    :param val: 
+    :param default: 
+
+    """
     return val if val is not None else default
 
 
 def cache(dir_path=None, enabled=None, verbosity=None, **cache_kwargs):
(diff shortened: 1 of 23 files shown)


0dm:
I should add that these docstrings don't really tell us anything new, they're just there..

abrichr:
Interesting, thank you @0dm ! At least it enforces a consistent style. This may be useful if we can't do it with pylint. Related: https://github.com/MLDSAI/OpenAdapt/pull/138

@Mustaballer any thoughts?

Mustaballer:
Sounds good, I think Google style is pretty clean and we can use it going forward

0dm:
@abrichr Yeah I agree, I didn't really want us to use it anyway - was just looking into it.

Issue #150, opened 2023-05-15
Add listener to stop recording


0dm:
We should add a listener for a key combination that stops recordings (default to ctrl+c, but preferably configurable) so that users don’t have to send SIGINT by interacting with the terminal. This would remove an unnecessary last-step of recordings.

I recommend we use pynput for this, since we already use it for other purposes - rough idea:
```py
# On Key Pressed Event                      
def on_press(key):
   ...

with kb.Listener(
        on_press=on_press) as listener:
    listener.join()
    ```

0dm:
It'd be best to run a check during recordings, while ActionEvents are being created.

abrichr:
Related: https://github.com/MLDSAI/OpenAdapt/pull/139

PR #200, opened 2023-05-29
Listener to stop recording

angelala3252:
This PR responds to [Issue #150](https://github.com/MLDSAI/OpenAdapt/issues/150). So far I've created a listener for the sequence "oa.stop" in the `record.py` file. Once this sequence is caught we do `terminate_event.set()`, exit the infinite loop for recording, and clean up the listener. I've briefly manually tested this and the recording does stop after typing "oa.stop".

Next steps:
- remove the combo from the recording (two different approaches suggested in https://github.com/MLDSAI/OpenAdapt/pull/139, so have to confirm which one to move forward with
- more thorough testing

Instructions to run:
```
git fetch angelala3252
git checkout feat/stop_listener
python -m openadapt.record "test" # stop recording using a stop sequence or Ctrl + C
```
Then when running, type any of the sequences in `config.STOP_SEQUENCE` to stop the recording.


openadapt/record.py
```diff
@@ -628,11 +628,44 @@ def record(
 
     # TODO: discard events until everything is ready
 
+    # Variables for sequence detection
+    sequence = ['o', 'a', '.', 's', 't', 'o', 'p']
```
abrichr, line 632: What do you think about moving this to `config.STOP_SEQUENCES` (a list of possible sequences)?

openadapt/record.py
```diff
@@ -628,11 +628,44 @@ def record(
...
+    current_index = 0
```
abrichr, line 633: What do you think about naming this more descriptively, e.g. `stop_sequence_index`.

In that case we may also need to maintain multiple indexes, one for each sequence in `config.STOP_SEQUENCES`.

openadapt/record.py
```diff
@@ -628,11 +628,44 @@ def record(
...
+    sequence_detected = False  # Flag to indicate if the sequence is detected
+
+    def on_press(key):
+        canonical_key = listener.canonical(key)[0]
+        nonlocal current_index, sequence_detected
+        sequence_canonical = listener.canonical(keyboard.KeyCode.from_char(sequence[current_index]))
```
abrichr, line 639: What do you think about using a consistent naming convention here? e.g. `canonical_sequence` to match `canonical_key`

openadapt/record.py
```diff
@@ -628,11 +628,44 @@ def record(
...
+        if canonical_key == sequence_canonical:
+            current_index += 1
+        else:
+            # Reset the index if the pressed key doesn't match the current key in the sequence
+            current_index = 0
+
+        # Check if the entire sequence has been entered correctly
+        if current_index == len(sequence):
+            # TODO: perhaps use logging to say we are stopping
+            print("Sequence entered correctly!")
```
abrichr, line 651: Yes, we should aways use logging unless there is a good reason not to.

abrichr (changes requested):
Thanks for getting started on this @angelala3252 !

angelala3252:
Update:

- I've implemented the changes Richard suggested above
- I've switched to using a list of stop sequences found at `config.STOP_SEQUENCES`, and adjusted the code to iterate through this list to check if any of these sequences have been entered. Any desired stop sequences can be added to this list. 
- Note: the stop sequences should be unique so that the user would most likely not enter this sequence for any other reason, the sequence is case insensitive (so oa.stop and OA.STOP will both be detected), and special characters/non-alphanumeric characters are currently not supported, (eg. control, enter...) since they are hard coded into pynput (keyboard.Key.ctrl) and you can't create a Key object from a special character like you can create a KeyCode object from an alphanumeric character. If we really want to be able to use special characters, I can see two solutions: either using pynput's HotKey functionality and requiring the special characters to be held down simultaneously, or adding the special characters to `config.STOP_SEQUENCES` as their keyboard.Key object.
- Ctrl + C can still be used to send KeyboardInterrupt to the terminal as usual, and should not be added to `config.STOP_SEQUENCES`, as any sequence in that list will terminate the recording regardless of which window is focused.

My next step is to filter out the termination sequence from the end of the recording once fetched, so I am thinking to make this change in `crud.get_latest_recording` for now.

0dm:
I'm trying to test what you currently have but I'm getting this issue:

```bat
2023-05-30 18:41:10.169 | INFO     | __main__:record:528 - task_description='test'
2023-05-30 18:41:10.206 | INFO     | __main__:create_recording:465 - recording=Recording(id=5, timestamp=1685486470.169697, monitor_width=4000, monitor_height=1080, double_click_interval_seconds=0.5, double_click_distance_pixels=5, platform='darwin', task_description='test')
2023-05-30 18:41:10.217 | INFO     | __main__:read_window_events:375 - starting
2023-05-30 18:41:10.221 | INFO     | __main__:read_screen_events:349 - starting
2023-05-30 18:41:10.224 | INFO     | __main__:process_events:72 - starting
zsh: abort      python3 -m openadapt.record "test"

OpenAdapt on  feat/stop_listener [$?] via 🐍 v3.10.11 (.venv) 
λ 

OpenAdapt on  feat/stop_listener [$?] via 🐍 v3.10.11 (.venv) 
λ 2023-05-30 18:41:11.185 | INFO     | __mp_main__:performance_stats_writer:422 - performance stats writer starting
2023-05-30 18:41:11.188 | INFO     | __mp_main__:write_events:211 - event_type='window' starting
2023-05-30 18:41:11.200 | INFO     | __mp_main__:write_events:211 - event_type='screen' starting
2023-05-30 18:41:11.233 | INFO     | __mp_main__:write_events:211 - event_type='action' starting
...
```

Not sure what's causing `zsh: abort      python3 -m openadapt.record "test"`, got any ideas?

Looks like the processes don't terminate either, gets a leak. 

Report:
```bat
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               Python [7139]
Path:                  /opt/homebrew/*/Python.framework/Versions/3.10/Resources/Python.app/Contents/MacOS/Python
Identifier:            org.python.python
Version:               3.10.11 (3.10.11)
Code Type:             ARM-64 (Native)
Parent Process:        zsh [1607]
Responsible:           Terminal [1593]
User ID:               501

Date/Time:             2023-05-30 18:46:27.1164 -0400
OS Version:            macOS 13.3 (22E252)
Report Version:        12
Anonymous UUID:        60EF07F9-F703-C1FC-ADD6-80A1F582223B


Time Awake Since Boot: 5700 seconds

System Integrity Protection: enabled

Crashed Thread:        12

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Application Specific Information:
ABORT -> Text Input Sources or Text Services Manager API is being called in two threads concurrently. If you are a UI application, you must call TIS/TSM API on the main thread. If you are a non-UI application (such as a cmd-line tool or a launchagent that does not run an event loop), you must not call TIS/TSM API from multiple threads concurrently.
abort() called


Thread 0::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x18aa1eff0 __select + 8
1   Python                        	       0x1031f0af4 time_sleep + 184
2   Python                        	       0x10308c524 cfunction_vectorcall_O + 100
3   Python                        	       0x103134c7c call_function + 128
4   Python                        	       0x103132494 _PyEval_EvalFrameDefault + 43104
5   Python                        	       0x1031269e0 _PyEval_Vector + 376
6   Python                        	       0x10313265c _PyEval_EvalFrameDefault + 43560
7   Python                        	       0x1031269e0 _PyEval_Vector + 376
8   Python                        	       0x103134c7c call_function + 128
9   Python                        	       0x103132544 _PyEval_EvalFrameDefault + 43280
10  Python                        	       0x1031269e0 _PyEval_Vector + 376
11  Python                        	       0x103134c7c call_function + 128
12  Python                        	       0x1031324bc _PyEval_EvalFrameDefault + 43144
13  Python                        	       0x1031269e0 _PyEval_Vector + 376
14  Python                        	       0x103134c7c call_function + 128
15  Python                        	       0x103132494 _PyEval_EvalFrameDefault + 43104
16  Python                        	       0x1031269e0 _PyEval_Vector + 376
17  Python                        	       0x103126854 PyEval_EvalCode + 104
18  Python                        	       0x1031216e4 builtin_exec + 380
19  Python                        	       0x10308c834 cfunction_vectorcall_FASTCALL + 88
20  Python                        	       0x103134c7c call_function + 128
21  Python                        	       0x1031324bc _PyEval_EvalFrameDefault + 43144
22  Python                        	       0x1031269e0 _PyEval_Vector + 376
23  Python                        	       0x103134c7c call_function + 128
24  Python                        	       0x1031324bc _PyEval_EvalFrameDefault + 43144
25  Python                        	       0x1031269e0 _PyEval_Vector + 376
26  Python                        	       0x1031a4038 pymain_run_module + 208
27  Python                        	       0x1031a3bd4 Py_RunMain + 980
28  Python                        	       0x1031a4f68 Py_BytesMain + 40
29  dyld                          	       0x18a6fbf28 start + 2236

Thread 1:
0   libsystem_pthread.dylib       	       0x18aa4ed8c start_wqthread + 0

Thread 2:
0   libsystem_pthread.dylib       	       0x18aa4ed8c start_wqthread + 0

Thread 3:
0   libsystem_kernel.dylib        	       0x18aa13f14 mach_msg2_trap + 8
1   libsystem_kernel.dylib        	       0x18aa26240 mach_msg2_internal + 80
2   libsystem_kernel.dylib        	       0x18aa1cb78 mach_msg_overwrite + 604
3   libsystem_kernel.dylib        	       0x18aa14290 mach_msg + 24
4   SkyLight                      	       0x18f82a75c SLWindowListCopyWindowInfo + 220
5   libffi.dylib                  	       0x19a91b050 ffi_call_SYSV + 80
6   libffi.dylib                  	       0x19a923af8 ffi_call_int + 1208
7   libffi.dylib                  	       0x19a923af8 ffi_call_int + 1208
```

angelala3252:
@0dm Weird, there seems to be an issue with time.sleep? Does recording work for you on the main branch?

0dm:
Yes

angelala3252:
@0dm `zsh abort` shouldn't be a problem with the code so I'm not sure what could be causing this. Are you running low on storage? I didn't update the requirements, but could you try running `pip install -r requirements.txt`?

0dm:
Shouldn't be storage related as everything works on the main branch

0dm:
I seemed to have fixed it, looks like 
```
Application Specific Information:
ABORT -> Text Input Sources or Text Services Manager API is being called in two threads concurrently. If you are a UI application, you must call TIS/TSM API on the main thread. If you are a non-UI application (such as a cmd-line tool or a launchagent that does not run an event loop), you must not call TIS/TSM API from multiple threads concurrently.
abort() called
```
was happening because I had Discord open in the background, even though I wasn't using it - which is quite weird.

I am getting another issue though, on keypress:
```
Unhandled exception in listener callback
Traceback (most recent call last):
  File "/Users/aaron/Documents/GitHub/OpenAdapt/.venv/lib/python3.10/site-packages/pynput/_util/__init__.py", line 230, in inner
    return f(self, *args, **kwargs)
  File "/Users/aaron/Documents/GitHub/OpenAdapt/.venv/lib/python3.10/site-packages/pynput/_util/darwin.py", line 286, in _handler
    self._handle(proxy, event_type, event, refcon, is_injected)
  File "/Users/aaron/Documents/GitHub/OpenAdapt/.venv/lib/python3.10/site-packages/pynput/keyboard/_darwin.py", line 262, in _handle
    key = self._event_to_key(event)
  File "/Users/aaron/Documents/GitHub/OpenAdapt/.venv/lib/python3.10/site-packages/pynput/keyboard/_darwin.py", line 322, in _event_to_key
    event_type = Quartz.CGEventGetType(event)
  File "/Users/aaron/Documents/GitHub/OpenAdapt/.venv/lib/python3.10/site-packages/objc/_lazyimport.py", line 168, in __getattr__
    value = getattr(p, name)
  File "/Users/aaron/Documents/GitHub/OpenAdapt/.venv/lib/python3.10/site-packages/objc/_lazyimport.py", line 186, in __getattr__
    value = self.__get_constant(name)
  File "/Users/aaron/Documents/GitHub/OpenAdapt/.venv/lib/python3.10/site-packages/objc/_lazyimport.py", line 412, in __get_constant
    self.__funcmap.pop(name)
KeyError: 'CGEventGetType'
```

angelala3252:
@0dm glad to hear the stop sequence is working at least. Does the keypress error only happen if you press a button before all the starting messages are logged? 

Seems the abort message could be because macOS doesn't like having two keyboard listeners at the same time, so I'll have to try to combine the two.

0dm:
Yes, only happens at the beginning.

angelala3252:
Update: 

- merged keyboard listeners in `record.py` for macOS compatability
- filter out Ctrl + C and any stop sequences in `config.STOP_SEQUENCES` in `crud.get_action_events`, so the stop sequence is saved to the database but whenever we retrieve the action events (visualizing, playback) it will not appear.

0dm:
Works great!

openadapt/config.py
```diff
@@ -84,6 +84,9 @@
     ],
 }
 
+# sequences that when typed, will stop the recording of ActionEvents in record.py
+STOP_SEQUENCES = [["o", "a", ".", "s", "t", "o", "p"]]
```
0dm, line 88: Do you think we should use strings here instead? We can then use .split() to achieve the same result, but it'd be easier + faster to define stop sequences this way.

0dm (commented):
lgtm well done

openadapt/record.py
```diff
@@ -466,12 +468,45 @@ def read_keyboard_events(
...
+        global sequence_detected
+
+        for i in range(0, len(config.STOP_SEQUENCES)):
+            # check each stop sequence
+            stop_sequence = config.STOP_SEQUENCES[i]
+            # stop_sequence_indices[i] is the index for this stop sequence
+            # get canonical KeyCode of current letter in this sequence
+            canonical_sequence = keyboard_listener.canonical(
+                keyboard.KeyCode.from_char(
+                    stop_sequence[stop_sequence_indices[i]]))
```
abrichr, line 495: @angelala3252 Can you please run `black`?

openadapt/record.py
```diff
@@ -466,12 +468,45 @@ def read_keyboard_events(
     terminate_event: multiprocessing.Event,
     recording_timestamp: float,
 ) -> None:
+
+      # create list of indices for sequence detection
+    stop_sequence_indices = []
+    for _ in config.STOP_SEQUENCES:
+        # one index for each stop sequence in config.STOP_SEQUENCES
+        stop_sequence_indices.append(0)
```
abrichr, line 476: How about:

```
stop_sequence_indices = [
    0 for _ in config.STOP_SEQUENCES
]
```

openadapt/record.py
```diff
@@ -466,12 +468,45 @@ def read_keyboard_events(
...
+        
     def on_press(event_q, key, injected):
         canonical_key = keyboard_listener.canonical(key)
         logger.debug(f"{key=} {injected=} {canonical_key=}")
         if not injected:
             handle_key(event_q, "press", key, canonical_key)
 
+        # stop sequence code
+        nonlocal stop_sequence_indices
```
abrichr, line 485: What do you think about storing this in

openadapt/config.py
abrichr: ```STOP_SEQUENCES = ["oa.stop".split()]```

openadapt/record.py
angelala3252: the default is 
```
            canonical_sequence = keyboard_listener.canonical(
                keyboard.KeyCode.from_char(stop_sequence[stop_sequence_indices[i]])
            )
```

Is this acceptable or would you prefer the format you wrote above?

openadapt/config.py
angelala3252: I've made this change and also fixed the listener code to properly support special characters, but the special characters in the stop strings shouldn't have the <> since I'm using the canonical names. Currently, I am splitting the strings by character, so I'm not sure how to split the ctrl c (or any other special characters).

openadapt/crud.py
```diff
@@ -109,36 +111,84 @@ def insert_recording(recording_data):
...
+    if len(action_events) >= 2:
+        if (
+            action_events[-1].canonical_key_char == "c"
+            and action_events[-2].canonical_key_name == "ctrl"
+        ):
+            # remove ctrl c
+            # ctrl c must be held down at same time, so no release event
+            action_events.pop()
+            action_events.pop()
+            return
```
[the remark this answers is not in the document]
angelala3252: added TODO in [90f95ad](https://github.com/MLDSAI/OpenAdapt/pull/200/commits/90f95ad9f899d6def98f61cc688140125cb7aa54)

abrichr:
@angelala3252 something's not quite right here.

If I create a recording by typing only "foo bar oa.stop" (with a pause between each key to make sure there is no overlap), the "bar' is removed.

If I perform the exact same actions but type "<ctrl>+c" to stop, the "bar" is not removed.

openadapt/config.py
abrichr: I think we would need to specify sequences more explicitly, e.g.:
```
STOP_STRS = [
    "oa.stop",
    ...
]
...
STOP_SEQUENCES = [
    stop_str.split() for stop_str in STOP_STRS
] + [
    ["<ctrl>+c"]
]
```

Note that `<ctrl>+c` denotes both pressed at the same time, whereas `<ctrl>c` would denote one pressed after the other.

Then we would need some extra logic to figure out when these sequences are triggered.

angelala3252:
@abrichr the visualization error should be fixed in [a2bd5ff](https://github.com/MLDSAI/OpenAdapt/pull/200/commits/a2bd5ff466f873fe1c610fccb6b8bdfc0e716189). 

as for your comment about how we specify stop sequences, I made some changes in [fa0d12a](https://github.com/MLDSAI/OpenAdapt/pull/200/commits/fa0d12aceec7bc662a16a50e91a1caef9dca04a1) that allow us to specify any stop sequences containing special characters, but they have to be specified key-by-key. Let me know if you think these changes are acceptable. This still doesn't differentiate between sequences that are held down at the same time and sequences that are pressed one key after the other, however, I do think having this functionality would be more trouble than it's worth, as I think that we should just try to use sequences that don't have any hotkey functionality so that it wouldn't matter if the keys are held down or pressed one after the other.

Issue #134, opened 2023-05-11
Create a GUI for OpenAdapt


0dm:
Story: As a user, I'd much rather prefer to have an interactive user interface instead of typing out commands for OpenAdapt. Using a terminal is too intimidating.

A GUI would be a great way to attract new users, and we should display buttons for Record / Stop, Visualize, and Replay.

PR #207, opened 2023-05-30
added openadapt.renew

0dm:
purpose: removes existing database file and runs alembic migrations
usage: `python3 -m openadapt.renew`

I find this pretty useful and like it more than typing `rm openadapt.db && alembic upgrade head` - and it's more pythonic..
I also added a definition for DB_FNAME because my linter was reporting an issue here:

image


.gitignore
```diff
@@ -21,4 +21,5 @@ cache
 performance
 
 # Generated when adding editable dependencies in requirements.txt (-e)
-src
\ No newline at end of file
+src
+.DS_Store
```
0dm, line 25: i always have to ignore the .DS_Store file on Mac 😅

openadapt/renew.py
```diff
@@ -0,0 +1,19 @@
+import os
+from subprocess import run, PIPE
+from openadapt.config import getenv_fallback
+
+
+def clear_db():
+    db = getenv_fallback("DB_FNAME")
+    if os.path.exists(db):
+        os.remove(db)
+    result = run(["alembic", "upgrade", "head"], stderr=PIPE, text=True)
```
0dm, line 10: prevents duplicate logging of config

abrichr: Please do!

0dm:
@abrichr ready again

openadapt/scripts/reset_db.py
```diff
@@ -0,0 +1,25 @@
...
+from subprocess import run, PIPE
+from openadapt.config import getenv_fallback
+
+
+def reset_db():
+    """
+    The function clears the database by removing the database file and running a
+    database migration using Alembic.
+    """
+    db = getenv_fallback("DB_FNAME")
```
abrichr, line 11: Why not just `config.DB_FNAME`?

openadapt/config.py
```diff
@@ -99,6 +99,7 @@ def getenv_fallback(var_name):
     locals()[key] = val
 
 ROOT_DIRPATH = pathlib.Path(__file__).parent.parent.resolve()
+DB_FNAME = getenv_fallback("DB_FNAME")
```
abrichr, line 102: Shouldn't this be defined already above? Can you please clarify why we are we explicitly redefining it here?

0dm: It's defined, but every linter keeps telling me it's undefined so it's mainly there just to silence the warnings:
image

0dm:
implemented in #134

0dm:
nevermind, there's a small change that needs to be merged (use config.DB_FPATH rather than getevv_fallback)

openadapt/config.py
abrichr: What do you think about just adding a `#noqa` directive (or similar) to suppress the warning?

KrishPatel13: Even I got this lint warning for DB_FNAME, but I always ignore it

openadapt/scripts/reset_db.py
```diff
@@ -1,16 +1,15 @@
...
 from subprocess import run, PIPE
-from openadapt.config import getenv_fallback
+from openadapt import config
 
 
 def reset_db():
     """
     The function clears the database by removing the database file and running a
     database migration using Alembic.
     """
```
KrishPatel13, line 10: A new line will be great as per the styling convention we are following.

0dm: Oh I was unaware of this since most styles are against newlines after method docstrings

openadapt/config.py
0dm: Yes, that worked.

Issue #37, opened 2023-04-27
Explore RWKV-LM


abrichr:
How can we use RWKV-LMM to implement "infinite" context lengths?

From https://github.com/BlinkDL/RWKV-LM:

> RWKV is an RNN with transformer-level LLM performance. It can be directly trained like a GPT (parallelizable). So it's combining the best of RNN and transformer - great performance, fast inference, saves VRAM, fast training, "infinite" ctx_len, and free sentence embedding.

Additional reading:

[From Deep to Long Learning?](https://hazyresearch.stanford.edu/blog/2023-03-27-long-learning)

[Hyena Hierarchy: Towards Larger Convolutional Language Models](https://arxiv.org/abs/2302.10866)

dianzrong:
RWKV-LM can be used throughout the project to:

- summarize texts associated with the screenshots
- better suggest prompts (like google's autofill)
- better understand prompts from the user
- in the future: handle translations

Implementation:
- can get "infinite" context lengths through iterative refinement
     - states would be the recording up until that point and the token would be the next screenshot
    
Limitation:
- this process can be computationally expensive

angelala3252:
As Dian mentioned above, RWKV-LM could be helpful for us since it could analyze large amounts of past user actions to better predict future actions. It could also learn how to do more complex tasks that the user themselves don't know how to do, as it could learn from longer texts like instruction manuals. These benefits would also help with efficiency and speed up the automation. 

As for implementation, we could consider combining RWKV-LM with another tool that could help with the graphical side of things. For example, we could have RWKV_LM understand the task we wish to execute and lay out the steps needed, then using something like OCR or MiniGPT-4, we could handle how to execute the steps on the GUI.

abrichr:
Closing as part of architecture transition. Legacy frozen at v0.46.0. See PR #960.

PR #209, opened 2023-06-01
Implementing RWKV-LLM (#37)

AvidEslami:
This pull requests implements RWKV's RNN model using Modal. (Issue #37 )

**Summary:** Created RWKV.py, (will move to openadapt/strategies/mixins and restructure to accept inputs)

Run using the following command (will be changed once implemented as a mixin):
```
modal run .\openadapt\RWKV\RWKV.py
``` 

**Implementation:**
 - Starts a Modal application, gpu must be set to a100 in order to run Raven-14B (largest model)
 - Downloads weights for the desired model from huggingface to the modal server
 - Computes and returns output

**TODO:** 
 - Change structure to function as a mixin (take inputs such as prompts or task descriptions)
 - Allow for modification of parameters via config file (Temperature, Top P, Presence Penalty, etc...)
 - Deploy Modal application permanently to avoid startup times (Upon startup weights must be downloaded, to avoid this we can deploy the Modal app externally and make requests to it instead)


requirements.txt
```diff
@@ -26,3 +26,5 @@ tiktoken==0.4.0
 torch==2.0.0
 tqdm==4.64.0
 transformers==4.28.1
+rwkv
+modal-client
```
abrichr, line 30: Please include versions

AvidEslami:
**Update:**
Run using the following commands 

first time using modal:
```
modal token new
```
then:
```
modal deploy .\openadapt\RWKV\RWKV.py
python .\openadapt\RWKV\test_RWKV.py
```
Modal server only needs to be deployed if the user wants to use modal (modify the "USE_MODAL" variable in config.py to run locally)

test_RWKV makes a simple call and just prints the output to test functionality.
To use RWKV as a mixin, refer to openadapt/strategies/mixins/rwkv.py

Note:
 - Startup takes a long time with modal since weights need to be downloaded to the modal server, repeated calls are slightly faster, but still slow
 - Models can be specified as follows from openadapt/strategies/mixins/rwkv.py
 ```Python
 # Choose a model:
MODEL = config.RWKV_MODEL
# model 0: RWKV-4-Raven-14B
# model 1: RWKV-4-Raven-7B
# model 2: RWKV-4-Pile-14B
# model 3: RWKV-4-World-1.5B
```
Model 0 is the most effective model in most cases.

**Comment by abrichr** [comment]
Thanks @AvidEslami ! What's the status on being able to run this locally? Can we add a `setup` method that automatically runs the required setup steps?

**Comment by AvidEslami** [comment]
> What's the status on being able to run this locally? Can we add a `setup` method that automatically runs the required setup steps?

Thanks for the comment @abrichr !
Commit [702e065](https://github.com/MLDSAI/OpenAdapt/pull/209/commits/702e065dbcd388787271269ce6ca43b28918d052) added a 'use_cuda' parameter to the function call which assists with running locally, I've tested it with the smallest model (model 2) and I was able to run it locally.

To test this, simply set USE_MODAL=False in config.py  , then select the desired model and run the following command:
```
python .\openadapt\RWKV\test_RWKV.py
```

When running the model locally for the first time, HuggingFace will download and cache the model weights, but no additional setup is required.

Unfortunately I don't think it would be possible to create a setup method for the Modal use-case since it requires that the user also creates an account on the Modal website.

I will resolve the merge conflicts, but please let me know if any other changes or desired or suggested!

Do you think I should add the following model descriptions to config.py? Currently they are located in openadapt/strategies/mixins/rwkv.py
```
# Choose a model:
MODEL = config.RWKV_MODEL
# model 0: RWKV-4-Raven-14B  <-- Largest model, finetuned version of model 3
# model 1: RWKV-4-Raven-7B
# model 2: RWKV-4-Raven-1B5 <-- smallest model
# model 3: RWKV-4-Pile-14B
# model 4: RWKV-4-World-1.5B
```

**Comment by abrichr** [comment]
@AvidEslami it would be nice to be able to run this with a CPU only. What do you think about https://github.com/saharNooby/rwkv.cpp ?

> Do you think I should add the following model descriptions to config.py? Currently they are located in openadapt/strategies/mixins/rwkv.py

Yes please!

**Comment by AvidEslami** [comment]
@abrichr 
Here are some test results I've gathered from various parameter setups, unfortunately the accuracy for most setups is fairly low, but please take a look and let me know if you have any suggestions for improving the tests, or creating new ones, thanks!
https://docs.google.com/document/d/13RJC1V_noQX08ZWpg10Ze8Yd1nhoCceGAWojUz2cBfI/edit?usp=sharing

**Comment by abrichr** [comment]
Thanks @AvidEslami ! I wonder if it might be more useful if we compress the prompt as much as possible, and remove any extraneous information. We should also try to remove any ambiguity (e.g. use "id" instead of "number", "descriptor" instead of "address", 0-based indexing). For example:

```
You are given a task, and a list of signals in JSON format. Please provide only the id of the signal that is most relevant to the task.

Task: filling forms and submitting data on web pages

Signals: 
[{'id': 0, 'type': 'file', 'descriptor': 'restaurant_menu_data.txt'}
,{'id': 1, 'type': 'url', 'descriptor': 'https://en.wikipedia.org/wiki/Web_development'}
,{'id': 2, 'type': 'function', 'descriptor': 'math.sqrt'}]

# Response:
```

AvidEslami:
> Thanks @AvidEslami ! I wonder if it might be more useful if we compress the prompt as much as possible, and remove any extraneous information.

Thanks for the advice @abrichr , I modified the setup which lead to more concise results but after running Raven-14B, and World-7B results were unfortunately still inaccurate:
https://docs.google.com/document/d/1xH1P2IiaKPME9gPvbueLv-PU66hkP3kLDtcE9gK6ws0/edit?usp=sharing

Please let me know if you have any other suggestions here!

abrichr:
As mentioned in Slack, I think it's time to start fine tuning 🤓

AvidEslami:
**Signals Finetune Update:**
 - The following sheet contains the outputs generated by the finetuned models [link](https://docs.google.com/spreadsheets/d/1JPY8nLpbvRKqnA9QgXwGOr5RarQ3oPYWFvLjDDA2sZs/edit?usp=sharing)
 - Any advice on the approach is appreciated, there are still several things to try:
   - Having a random numbers of signals, which could perhaps reduce the models chance of memorizing results
   - Changing prompt structure could improve understanding
   - Creating a more varied dataset (more signals / more tasks) help model catch relations
   - Finetuning Pile-14B for several epochs (will try), Pile-14B is the non finetuned version of Raven-14B
  
As usual please let me know if you have any suggestions!

abrichr:

Issue #27, opened 2023-04-19
Implement Gisting


abrichr:
How can we take advantage of [gisting](https://github.com/jayelm/gisting)?

[Thread](https://twitter.com/jayelmnop/status/1648365743195684873)

[Paper](https://arxiv.org/abs/2304.08467)

FFFiend:
I've tried mapping the concept to fit Puterbot's case but I'm having trouble understanding what the paper is essentially proposing. Are we compressing prompts into "gist tokens" after initially feeding the LM the entire uncompressed prompt so that it saves on read and compute time later on? Or are we compressing it from the beginning ? The attention masks part is especially confusing, so I was wondering if I could get someone else's perspective on Gisting :D

abrichr:
Thank you for engaging on this issue @FFFiend !

> Are we compressing prompts into "gist tokens" after initially feeding the LM the entire uncompressed prompt so that it saves on read and compute time later on?

I believe that is correct. From https://arxiv.org/pdf/2304.08467.pdf:

> we add a single special gist token to
> the model vocabulary and embedding matrix, much
> like the start/end-of-sentence tokens often present
> in such models. Then, given a (task, input) pair
> (t, x), we concatenate t and x with a set of k successive gist tokens in between: (t, g1, . . . , gk, x), e.g.
> Translate French: <G1> <G2> The cat. This
> sequence is fed into the model, with the restriction that input tokens after the gist tokens cannot
> attend to any of the prompt tokens before the gist
> tokens (but they can attend to the gist tokens). This
> forces the model to compress the information in
> the prompt into the gist tokens, since the input x
> (and output y) cannot attend to the prompt t.

According to the repo (https://github.com/jayelm/gisting) it appears that the trained gist model parameter diffs are now available on HuggingFace for LLaMA-7B and FLAN-T5-XXL 😄

FFFiend:
Alright so I think I spent an unreasonable amount of time trying to understand how exactly the gist tokens themselves looked like from multiple reads on the paper, but I think the repo you linked above blackboxes it for us. 

What I think remains now is to take advantage of the above repo such that any prompting we do on the LLM we use to automate processes, can be gisted and we save on compute time. Unfortunately, I haven't tinkered around with the Puterbot codebase enough yet to try and see where we can inject this improvement :/ Will comment/commit soon :D

abrichr:
[...]

PR #222, opened 2023-06-05
Gisting

FFFiend:
Addresses #27 , please run ```gist_test.py``` using Modal to view OOM error.

I've highlighted this in the actual docstring, however you can go into the ```compress.py``` file inside where you installed the module (lines 105-107, device_map="auto" and nn.Parallel(..)) , and observe how I've tried enabling multi GPU usage, to no avail :(


FFFiend:
Ready for review @abrichr or rather investigation, because it refuses to run in parallel across the GPU's.

abrichr:
Thanks @FFFiend ! Can you please clarify where `gisting_test` is defined? Ideally we would also have a script to deploy this automatically.

abrichr:
Please add links to all of your conversations with third parties here 🙏

abrichr:
Please spec out an AWS instance that would run Gisting 🙏

abrichr:
@FFFiend can you please suggest the minimum required specifications for an AWS instance we can deploy to run this?

FFFiend:
According to the comment here https://github.com/jayelm/gisting/issues/4, it should be possible to run on a single A100 40 GB GPU but given that that's not the case, the absolute minimum I can think of is 2 A100 40GB GPU's, or perhaps just one singular 80 GB GPU if that's available. This would require a one line modification in ```compress.py```'s code though, and that is adding ```device_map="auto"``` into the model initializer.

abrichr:
Moved to https://github.com/OpenAdaptAI/OpenAdapt/pull/395/files

Issue #141, opened 2023-05-11
Implement MPT-7B


atomicrichard:
How can we take advantage of [MPT-7B](https://old.reddit.com/r/MachineLearning/comments/138sdwu/n_introducing_mpt7b_a_new_standard_for_opensource/)?

abrichr:
Closing as part of architecture transition. Legacy frozen at v0.46.0. See PR #960.

PR #226, opened 2023-06-06
MPT-7B Implementation

FFFiend:
Addresses #141 . 

Completions seem straight up incorrect and arbitrary, sometimes using different languages to complete the sentence LOL. This is probably due to the base model not being fine tuned on a bunch of English text before they pushed it to HF.

Here are some examples of completions. Completion length also seems super short, although there should be some parameter in the MPT-7B docs that we can use to change this?

Prompt: "Tom and Jerry "
Screen Shot 2023-06-13 at 5 08 33 PM

Prompt: "Melbourne is a "
Screen Shot 2023-06-13 at 5 08 52 PM

Lastly, this is run through Modal, and I know mixins arent meant to be run directly, so I'm confused on the final state of the implementation you'd like this to be in. If we use a regular class, then we'd have to call the ```get_completions``` method from a model entry point.


abrichr:
@FFFiend thank you! What might this be useful for?

FFFiend:
Since it's open source, MPT-7B is a potential candidate for training Event generation and completions on. The only concerning part is the unusually short response length.

abrichr:
Thanks @FFFiend ! Let's try running this offline on CPU

FFFiend:
> Thanks @FFFiend ! Let's try running this offline on CPU

Done! See here: https://github.com/MLDSAI/OpenAdapt/pull/303

abrichr:
@FFFiend should we close this in favor of https://github.com/OpenAdaptAI/OpenAdapt/pull/303 ?

FFFiend:
Absolutely, the GGML model can actually be run on Macs unlike this one :)

Issue #170, opened 2023-05-25
Implement tracing


abrichr:
Simplify debugging: https://blog.koehntopp.info/2023/03/14/tracing-python.html

abrichr:
From https://blog.koehntopp.info/2023/03/14/tracing-python.html :

```
import functools


def args_to_str(*args):
    return ", ".join(map(str, args))


def kwargs_to_str(**kwargs):
    ret = ""
    for k, v in kwargs.items():
        ret += f"{k}={v},"

    return ret[:-1]


def logging(func):
    func.__indent__ = 0

    @functools.wraps(func)
    def wrapper_logging(*args, **kwargs):
        func_indent = " " * func.__indent__
        func.__indent__ += 2

        func_name = func.__qualname__
        func_args = args_to_str(*args)
        func_kwargs = kwargs_to_str(**kwargs)

        print(f"{func_indent} -> Enter: {func_name}({func_args}", end="")
        if func_kwargs != "":
            print(f", {func_kwargs}", end="")
        print(")")

        result = func(*args, **kwargs)

        print(f"{func_indent} <- Leave: {func_name}({result})")
        return result

    return wrapper_logging


@logging
def fac(n: int) -> int:
    if n == 1:
        return 1
    else:
        return n * fac(n - 1)


if __name__ == '__main__':
    result = fac(3)
    print(f"{result=}")
```

PR #229, opened 2023-06-07
Implement tracing

angelala3252:
This PR addresses [issue #170 ](https://github.com/MLDSAI/OpenAdapt/issues/170) by adding the @logging decorator shown in [this article](https://blog.koehntopp.info/2023/03/14/tracing-python.html). I changed the print statements to logging as it got a little messy with print statements alternating with the logging, and I also got rid of the indentation as it didn't seem necessary when using logging. I didn't add the @logging decorator to the functions that got triggered a lot like the event handlers. 

Next steps:
- the `record` function still hangs sometimes near the end without any additional info from the added tracing, so I will have to find another way to give more information about what is causing the issue
- even without the tracing, all the logging from different threads often gets mixed together into one sentence (example shown in the image below). If we want to make the logging more readable, we will have to make sure that only one thread logs at one time.

image

Instructions to run:
``` 
git fetch angelala3252
git checkout feat/tracing
python -m openadapt.record "test" 
# tracing will show up in the terminal, "Enter" for when you enter a function, and "Leave" for when a function returns
```


angelala3252:
Update:
The log messages now no longer get mixed up as they are now multiprocess safe. The recording hanging issue was due to a separate error I was getting, so now the recording does not hang and every major function gets traced. I haven't gone any further with the tracing itself, as it still just shows entering and exiting functions, but if that is necessary I can explore that more.

abrichr:
@angelala3252 can you please fix merge conflicts?

angelala3252:
@abrichr resolved!

openadapt/record.py
```diff
@@ -37,6 +39,43 @@
...
+
+
+def args_to_str(*args):
+    return ", ".join(map(str, args))
+
+
+def kwargs_to_str(**kwargs):
+    ret = ""
+    for k, v in kwargs.items():
+        ret += f"{k}={v},"
```
abrichr, line 52: What do you think about:
```
return ",".join([f"{k}={v}" for k, v in kwargs.items()])
```
🤓

openadapt/record.py
```diff
@@ -37,6 +39,43 @@
...
+
+    return ret[:-1]
+
+
+def logging(logger):
```
abrichr, line 57: Maybe we should name this differently, e.g. `trace(logger)`, what do you think?

openadapt/utils.py
```diff
@@ -26,9 +27,16 @@ def configure_logging(logger, log_level):
     log_level_override = os.getenv("LOG_LEVEL")
     log_level = log_level_override or log_level
     logger.remove()
+    logger_format = ("<green>{time:YYYY-MM-DD HH:mm:ss.SSS}</green> | "
+                     "<level>{level: <8}</level> | "
+                     "<cyan>{name}</cyan>:<cyan>{function}</cyan>:<cyan>{line}</cyan> "
+                     "- <level>{message}</level>")
```
abrichr, line 33: Can you please format this as follows:
```
    logger_format = (
        "<green>{time:YYYY-MM-DD HH:mm:ss.SSS}</green> | "
         "<level>{level: <8}</level> | "
         "<cyan>{name}</cyan>:<cyan>{function}</cyan>:<cyan>{line}</cyan> "
         "- <level>{message}</level>"
)
```
Also can you please clarify the origins of this format?

openadapt/record.py
angelala3252: renamed in [db24c49](https://github.com/MLDSAI/OpenAdapt/pull/229/commits/db24c49d6d83631b825b6706c3efa316f9ac5030)

PR #230, opened 2023-06-07
feat(opeadapt): add native chrome extension

KrishPatel13:
Progress:

The extension currently sends the documentInfo object: 
```
const documentInfo = {
    title: document.title,
    url: window.location.href,
    html: document.documentElement.outerHTML
};
```
On every change in the DOM. 

Example of how does the console receive messages:
image
so on


abrichr:
Good work @KrishPatel13 !

Next up is connecting the extension to python and having the python module the updates.

KrishPatel13:
@abrichr, Just to be clear, Currently I have a python module, browser.py that gets the updates in DOM and then it sends the reply_num (update count) and the whole documentInfo object via native messaging to the background.js script which logs in the console.

See here: 
https://github.com/MLDSAI/OpenAdapt/pull/230/files#diff-c51a7447dc4c4bc4b75e74075f2d00cdf140a026fe7dc02e6de6d64fcec06be5R44

My screenshot here is the exact description of the browser.py in action with background and content script :)
https://github.com/MLDSAI/OpenAdapt/pull/230#issue-1746148482

KrishPatel13:
> Refer

From this, I interpret as, making a new python module that sends and then logs only the updated part of the DOM. 
Is that right ? :)

abrichr:
https://learn.microsoft.com/en-us/windows/win32/ipc/pipes

abrichr:
Super early but looks very relevant

KrishPatel13:
@abrichr 

On researching and reading about pipes, I found that:

Creating a platform-independent solution that uses named pipes can be challenging, because named pipes work differently on Unix/Linux versus Windows systems. So essentially we would need different scripts for Windows and Unix/Linux and/or macOs.

A more platform-independent approach might be to use Python's built-in support for inter-process communication (IPC) via sockets. This approach works similarly to pipes but should function the same way across different operating systems. 

Or

Another way I can think of is to make this work by writing into a temp txt file. And then we want to run openadapt, we can get the data from it. If this works, we can measure the performance difference. If is huge, we can work on improving the performance. But, at least we would have something working the end :)


Please let me know your thoughts. 
THank you.

abrichr:
Please remove the `.idea` directory 🙏

KrishPatel13:
A Demo: https://mldsai.slack.com/archives/D056HTS5KEG/p1687015265698859

KrishPatel13:
Next Steps:

- [ ] Integrate the named-piped scripts with browser extension to print the actual dom changes in extension
- [ ] Integrate the whole environment with openadapt.

native_chrome_extension/content.js
```diff
@@ -0,0 +1,24 @@
...
+ * Send the document info to the background script
+*/
+function sendDocumentInfo() {
+    const documentInfo = extractDocumentInfo();
+    chrome.runtime.sendMessage({ type: 'document', documentInfo });
+}
+
+
+sendDocumentInfo();
+document.addEventListener('DOMSubtreeModified', sendDocumentInfo);
```
KrishPatel13, line 24: Alternative to Mutation Observer but use DOM Subtree Modified.

abrichr: Can you please list tradeoffs between these approaches?

native_chrome_extension/server.py
```diff
@@ -0,0 +1,37 @@
...
+        return 'quit'
+
+
+def server():
+    with NPopen('wt', name='test') as pipe:  # writable text pipe
+        print('Pipe Created = ', pipe.path)
+    
+        print('waiting for client to connect...')
+        stream = pipe.wait()
+        print('client connected')
```
abrichr, line 19: Please use logger.info 🙏

KrishPatel13:
Closing in favour of: https://github.com/OpenAdaptAI/OpenAdapt/pull/337

Issue #173, opened 2023-05-26
Explore productivity measurement


abrichr:
How can we measure productivity in a recording, both that of a user and of an agent?

angelala3252:
Our definition of productivity would probably relate to both efficiency and whether the user is actively doing something most of the time in a recording. Things that would be considered unproductive are redundant steps, long pauses (gaps between ActionEvent timestamps), making many errors, steps that take a lot of time that could be optimized, excessive mouse movements, or unnecessary keystrokes. Ideally, we want to identify all these things in a recording that could result in wasting time and minimize the amount of time wasted. This could also help demos be more compelling and show how OpenAdapt can automate tasks to be much more efficient. 

Traditional methods of measuring productivity might not necessarily work for our purposes but could still be useful to think about:
- counting the # of tasks completed, and measuring the amount of time for each task: hard to compare different tasks as they might not be the same difficulty
- input/output ratio: eg. comparing money produced to amount of time, 10000$ produced in 62 hours -> 10000/62 ~= 160$ produced per hour. For our case, it's unclear what general input and output we would measure would be

Some approaches we can consider:
- compare how long the user takes to do a task, and how long OpenAdapt takes to do it
- similarly, we could compare the # of ActionEvents needed

abrichr:
What are some interesting ways we could visualize this information and present it to business leadership?

e.g. infinite canvas displaying graph of condensed trees similar to visualize.py in which each node is a separate workflow cluster

angelala3252:
I think a good starting point would be having pictures of the windows on the left and then on the right, productivity info. Above everything we can have overall info about the whole recording. This would be very similar to visualize.py. For now, I'll just have basic info like time spent on each window, and then we can definitely add more later and make it look prettier if needed.

Is this similar to what you're imagining?

Jonathan-Sendsen:
Going to try and answer a few [slack questions](https://mldsai.slack.com/archives/C050869D3K4/p1685578540470109) here 😅 :

**Understanding the first use case we should prioritize:**

Given our objectives, customer problems identified and value props the first use case should ideally be something that is repetitive, time-consuming, prone to human error, and has clear steps that can be automated. The ideal use case should also be something that is frequent enough to provide significant time savings if automated.

A good example might be data entry or administrative tasks. Such tasks often have specific, step-by-step procedures, they're time-consuming, and there's a high probability of human error.

**Measuring productivity in a recording:**

Here are a few suggestions:

[...]

PR #234, opened 2023-06-08
Productivity measurement

angelala3252:
This PR responds to [issue #173](https://github.com/MLDSAI/OpenAdapt/issues/173) by gathering basic productivity information from the most recent recording and displaying it on an HTML page (similar to `visualize.py`). The screenshots on the left are those where a window event occurs (changing tabs, changing windows, etc.), and there is some basic information displayed to the right of that screenshot about what was done on that window (# of clicks, # of keys...).

Ex:
Overall info at the top
image
1st window
image
2nd window
image

Instructions to run:
```
git fetch angelala3252
git checkout feat/productivity
python -m openadapt.productivity 
```
For a better example try to switch tabs and do multiple actions on each tab, and try pausing for a bit (you can change the number of seconds that gets recognized as a long pause at `productivity.MAX_GAP_SECONDS`.


angelala3252:
Updates:

Added some info to the overview section
image
And added time spent on each window (although the window-by-window section has some issues and isn't 100% accurate)
image

Now working on identifying repetitive tasks from the `ActionEvents` and making it more accurate. I've tried using https://github.com/oemof/cydets but it identifies almost everything as a cycle, even if it doesn't repeat and has no pattern to it. I've tried a brute force method but it was way too slow. I will continue trying to optimize my brute force method, as well as researching to see if there are any algorithms I can use.

angelala3252:
I've implemented Brent's algorithm to try to find cycles in the ActionEvents. I tested it on recordings containing repetitions of the simple task of 1. click start 2. click word 3. click start 4. click settings. In a recording containing two back-to-back repetitions of the simple task, it found 2 repetitions. In another recording containing two repetitions of the simple task separated by an unrelated click, it also found 2 repetitions. I will have to do more testing on different examples, but I think this is a great result so far.

abrichr:
@angelala3252 any thoughts on https://huggingface.co/blog/autoformer ?

> Autoformer builds upon the traditional method of decomposing time series into seasonality and trend-cycle components. This is achieved through the incorporation of a Decomposition Layer, which enhances the model's ability to capture these components accurately. Moreover, Autoformer introduces an innovative auto-correlation mechanism that replaces the standard self-attention used in the vanilla transformer. This mechanism enables the model to utilize period-based dependencies in the attention, thus improving the overall performance.

abrichr:
@angelala3252 another way to think about this would be to measure the "similarity" between sequences, rather than identifying equality.

e.g. https://arxiv.org/abs/2306.05426

angelala3252:
Using the longest repeated non-overlapping substring algorithm found [here](https://www.geeksforgeeks.org/longest-repeating-and-non-overlapping-substring/), I am now able to identify any task that repeats in a recording, whether the tasks are back to back or not. 🥳

angelala3252:
Currently, the task identification has inconsistent/unpredictable behaviour on longer recordings (the identified task contains the current task and part of the next task), so I'm working on trying to fix this issue.

angelala3252:
Example pdf: 
[Productivity metrics for recording-28.pdf](https://github.com/OpenAdaptAI/OpenAdapt/files/12264540/Productivity.metrics.for.recording-28.pdf)

openadapt/productivity.py
```diff
@@ -0,0 +1,674 @@
...
+
+    $ python openadapt/productivity.py"
+"""
+
+from pprint import pformat
+from threading import Timer
+import os
+import string
+from typing import Optional, Tuple
+
```
KrishPatel13, line 15: The order of imports is incorrect here.

Please see Contributing md file for details

KrishPatel13 (commented):
@angelala3252 I would strong encourage you to run:

`black <on all of the changed python files>`
`flake8 <on all of the changed python files>`
and then fix all errors if any

Optinoal but I do it:
`pylint <on all of the changed python files>`

openadapt/productivity.py
KrishPatel13: Other flake8 issues:
https://github.com/OpenAdaptAI/OpenAdapt/pull/474#issuecomment-1692673244

angelala3252:
@KrishPatel13 I believe I finished this PR before I installed the pre-commit which is why I missed some flake8 errors previously 😅  This new PR should resolve the errors
https://github.com/OpenAdaptAI/OpenAdapt/pull/480

abrichr:
Thank you for the great work @angelala3252 !

Regarding your comment on Aug 4:

> Example pdf:
> [Productivity metrics for recording-28.pdf](https://github.com/OpenAdaptAI/OpenAdapt/files/12264540/Productivity.metrics.for.recording-28.pdf)

This appears to have the productivity metrics cut off or otherwise ineligible:

image

Would be great to polish this off so we can show it off on the README 😄

PR #307, opened 2023-06-22
add python-Levenshtein to poetry

0dm:
simple addition to resolve the following warning (seen after pytest):

```python
UserWarning: Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning
    warnings.warn('Using slow pure-python SequenceMatcher. Install python-Levenshtein to remove this warning')
```

What is Levenshtein?:
The Levenshtein Python C extension module contains functions for fast computation of:
- Levenshtein (edit) distance, and edit operations
- string similarity
- approximate median strings, and generally string averaging
- string sequence and set similarity

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1766,6 +1766,120 @@ files = [
 ]
 
 [[package]]
+name = "levenshtein"
+version = "0.21.1"
+description = "Python extension for computing string edit distances and similarities."
+optional = false
+python-versions = ">=3.6"
+files = [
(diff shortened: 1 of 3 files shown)


dianzrong:
Thanks for adding it!! Could you also add it to requirements just in case someone wants to use it?

0dm:
> Thanks for adding it!! Could you also add it to requirements just in case someone wants to use it?

Added 👍

KrishPatel13 (approved):
Reviewed!

dianzrong (approved):
Looks good to me 😻

0dm:
@abrichr here's one we merged without you

Issue #187, opened 2023-05-27
Implement layout extraction


abrichr:
In order to support https://github.com/MLDSAI/OpenAdapt/issues/157, we want to extract structured information from documents.

See https://huggingface.co/docs/transformers/model_doc/layoutlm for implementation (alternatives?)

FFFiend:
for the ```image``` param would it be called using the ```image``` property in ```Screenshots```?

abrichr:
Closing as part of architecture transition. Legacy frozen at v0.46.0. See PR #960.

PR #341, opened 2023-07-03
LayoutExtractionReplayStrategyMixin

FFFiend:
Adds a replay strategy to extract key information from text-based documents using document-question answering. 

**Summary**
addresses #187.

**Checklist**
* [x] If applicable, I have added tests to prove my fix is functional/effective


openadapt/strategies/mixins/layout.py
```diff
@@ -0,0 +1,36 @@
...
+            for img_file_path in image_file_paths
+        ]
+
+    def document_query(self, image: np.ndarray, question: str) -> str:
+        query_pipeline = pipeline(
+            "document-question-answering", 
+            model="impira/layoutlm-document-qa"
+        )
+        output = query_pipeline(image, question)
+        if output != []:
```
abrichr, line 34: How about `if not output`?

openadapt/strategies/mixins/layout.py
```diff
@@ -0,0 +1,36 @@
...
+            return query_pipeline(image, question)[0]['answer']
+        return "Unsupported document type, please input a text-based document"
```
abrichr, line 36: What do you think about raising an exception here?

tests/openadapt/test_layout.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class TestQAPipeline:
+    def __init__(self, image_file_paths: List[str]):
+        self.image_list = [
+            Image.open(img_file_path).convert("RGB")
+            for img_file_path in image_file_paths
+        ]
+
+    def document_query(self, image: np.ndarray, question: str) -> str:
```
abrichr, line 15: Why not test `LayoutExtractionReplayStrategyMixin.document_query` directly? It seems we are duplicating functionality here.

tests/openadapt/test_layout.py
```diff
@@ -0,0 +1,97 @@
...
+        query_pipeline = pipeline(
+            "document-question-answering", model="impira/layoutlm-document-qa"
+        )
+        output = query_pipeline(image, question)
+        if output != []:
+            return query_pipeline(image, question)[0]["answer"]
+        return "Unsupported document type, please input a text-based document"
+
+
+IMAGE_LIST = [
```
abrichr, line 25: This seems like it should be called `IMAGE_FILE_NAMES`

openadapt/strategies/mixins/layout.py
```diff
@@ -0,0 +1,36 @@
...
+
+from openadapt.models import Recording, Screenshot
+from openadapt.strategies.base import BaseReplayStrategy
+
+
+class LayoutExtractionReplayStrategyMixin(BaseReplayStrategy):
+
+    def __init__(self, recording: Recording, image_file_paths: List[str]):
+        super.__init__(recording)
+        self.image_list = [
```
abrichr, line 23: What do you think about naming this `images`?

tests/openadapt/test_layout.py
```diff
@@ -0,0 +1,97 @@
...
+def test_invoice_screenshot():
+    questions = [
+        "What is the sender's phone number?",
+        "What is the sender's address?",
+        "What is surcharge?",
+        "What is the total cost?",
+        "What is the daily interest rate on overdue balances?",
+    ]
+
+    answers = ["321-956-7331,", "MELBOURNE,", "45.00", "$ 3,165.00", "0.5%"]
```
abrichr, line 66: `expected_answers`

tests/openadapt/test_layout.py
```diff
@@ -0,0 +1,97 @@
...
+        output = LAYOUT_OBJ.document_query(image=LAYOUT_OBJ.image_list[1], question=q)
+        output_answers.append(output)
+    assert output_answers == answers
+
+
+def test_calendar_screenshot():
+    output = LAYOUT_OBJ.document_query(
+        image=LAYOUT_OBJ.image_list[2], question="What month is it?"
+    )
+    assert output == "June" or output == "JUNE"
```
abrichr, line 79: `output.lower() == 'june'`

tests/openadapt/test_layout.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def test_calc_screenshot():
+    output = LAYOUT_OBJ.document_query(
+        image=LAYOUT_OBJ.image_list[3],
+        question="What is the current number on the screen?",
+    )
+    assert (
+        output == "10"
+        or output == "Unsupported document type, please input a text-based document"
```
abrichr, line 89: This will have to change if we raise an exception

abrichr (changes requested):
Looks good @FFFiend ! Just a few small clarifications

openadapt/strategies/mixins/layout.py
FFFiend: Good point, I was using ```if output is not None``` initially but I remembered from way back in first year that there's a difference between ```is not ``` and ```==/!=``` but I didn't think of using the boolean evaluation. I'll make this change 👍

tests/openadapt/test_layout.py
FFFiend: ```BaseReplayStrategy``` has an abstract method, get_next_action_event and so it threw an error whenever I tried to use the class because I didn't implement it in ```LayoutExtractionReplayStrategyMixin```..not sure why I didn't just implement and return a call to the base class's method in the first place but I've added the necessary changes to get this working 👍

openadapt/strategies/mixins/layout.py
FFFiend: actually I tried this change out just now, and it still throws the ```IndexError: list index out of range``` whenever it returns an empty list (i.e. when the model has no answer to provide due to an unsupported document, in case of the calculator GUI screenshot for example) and so keeping it as ```if output != []``` is better since we can then raise a ```TypeError: Unsupported document...``` error

abrichr (changes requested):
Thanks @FFFiend ! Can you please move the images to a new directory at `tests/assets`? Also please move 
`assets/test_scrub_image.png` to this directory.

openadapt/strategies/mixins/layout.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def document_query(self, image: np.ndarray, question: str) -> str:
+        query_pipeline = pipeline(
+            "document-question-answering", 
+            model="impira/layoutlm-document-qa"
+        )
+        output = query_pipeline(image, question)
+        if not output:
+            raise TypeError("Unsupported document type, please input a text-based document")
+        return query_pipeline(image, question)[0]['answer']
```
abrichr, line 36: Should this be:
```
output[0]['answer']
```

openadapt/strategies/mixins/layout.py
```diff
@@ -0,0 +1,39 @@
...
+    
+    def get_next_action_event(self, screenshot: Screenshot) -> ActionEvent:
```
abrichr, line 38: Can you please remove this?

tests/openadapt/test_layout.py
```diff
@@ -0,0 +1,78 @@
...
+import pytest
+
+IMAGE_FILE_NAMES = [
+    "assets/test_ladingbill.png",
+    "assets/test_invoice.png",
+    "assets/test_calendar.png",
+    "assets/test_calc.png",
+]
+
+LAYOUT_OBJ = LayoutExtractionReplayStrategyMixin(Recording(), IMAGE_FILE_NAMES)
```
abrichr, line 16: Rename to `replay_strategy`

openadapt/strategies/mixins/layout.py
```diff
@@ -0,0 +1,39 @@
...
+class LayoutExtractionReplayStrategyMixin(BaseReplayStrategy):
+
+    def __init__(self, recording: Recording, image_file_paths: List[str]):
+        super().__init__(recording)
+        self.images = [
+            Image.open(img_file_path).convert("RGB")
+            for img_file_path in image_file_paths
+        ]
+
+    def document_query(self, image: np.ndarray, question: str) -> str:
```
abrichr, line 28: Let's move this function to `openadapt/tools/layout.py`. Let's also modify it as follows:
```
def document_query(self, image: np.ndarray | None, image_path: str | None, question: str) -> str:
    assert image or image_path, "At least one of image or image_path must be supplied"
    assert not (image and image_path), "Only one of image or image_path may be supplied"
    if image_path:
        image = Image.open(image_path).convert("RGB")
    ...
```

tests/openadapt/test_layout.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+    assert output.lower() == "june"
+
+
+def test_calc_screenshot():
+    output = LAYOUT_OBJ.document_query(
+        image=LAYOUT_OBJ.images[3],
+        question="What is the current number on the screen?",
+    )
+    assert(output==10 or pytest.raises(TypeError))
```
abrichr, line 71:
```
EXPECTED_OUTPUT = 10
assert output == EXPECTED_OUTPUT
```

We should never have magic numbers: https://en.wikipedia.org/wiki/Magic_number_(programming)

openadapt/tools/layout.py
```diff
@@ -0,0 +1,32 @@
+from transformers import pipeline
+from PIL import Image
+from typing import Optional
+import numpy as np
+
+
+class LayoutExtractionTool:
+    def document_query(
```
FFFiend, line 8: no data is being stored within the class and so the class has no attributes, just a directly callable method.

abrichr: No class needed, can you please make this a function?

openadapt/tools/layout.py
```diff
@@ -0,0 +1,32 @@
...
+            document_image = image
+
+        query_pipeline = pipeline(
+            "document-question-answering", model="impira/layoutlm-document-qa"
+        )
+
+        output = query_pipeline(document_image, question)
+        if not output:
+            raise TypeError(
+                "Unsupported document type, please input a text-based document"
```
abrichr, line 30: This assumes that the model is infallible, which we know is not the case.

How about simply returning `output` and letting the caller decide how to handle it?

FFFiend: Sounds good to me, although I would like to raise a warning of some kind to use this exclusively on text based documents.

abrichr: @FFFiend how about:

```
output = query_pipeline(document_image, question)
if not output:
    logger.warning("{output=}")
return output
```

openadapt/tools/layout.py
```diff
@@ -0,0 +1,27 @@
...
+        document_image = image
+
+    query_pipeline = pipeline(
+        "document-question-answering", model="impira/layoutlm-document-qa"
+    )
+
+    output = query_pipeline(document_image, question)
+    if not output:
+        logger.warning("Answer may be inaccurate as a non text-based file has been used")
+        return output
```
abrichr, line 26: Remove

openadapt/tools/layout.py
```diff
@@ -0,0 +1,27 @@
...
+    else:
+        document_image = image
+
+    query_pipeline = pipeline(
+        "document-question-answering", model="impira/layoutlm-document-qa"
+    )
+
+    output = query_pipeline(document_image, question)
+    if not output:
+        logger.warning("Answer may be inaccurate as a non text-based file has been used")
```
abrichr, line 25: This warning message is inaccurate. It is possible that the model simply failed.

openadapt/tools/layout.py
```diff
@@ -0,0 +1,27 @@
...
+        return output
+    return output[0]["answer"]
```
abrichr, line 27: This should be:

```
    output = query_pipeline(document_image, question)
    if not output:
        logger.warning(f"{output=}")
    return output
```

abrichr, line 27:
```suggestion
    if not output:
        logger.warning(f"{output=}")
    return output
```

tests/openadapt/test_layout.py
```diff
@@ -0,0 +1,71 @@
...
+        "What is the bill of lading number?",
+        "What is the carrier name?",
+        "What is the ship date?",
+        "What is the sender's phone number?",
+    ]
+
+    output_answers = []
+    for q in questions:
+        output = document_query(image=None, image_path=IMAGE_FILE_NAMES[0], question=q)
+        output_answers.append(output)
```
abrichr, line 27:
```suggestion
    output_answers = []
    for q in questions:
        output = document_query(image=None, image_path=IMAGE_FILE_NAMES[0], question=q)
        output_answers.append(output[0]["answer"])
```

openadapt/tools/layout.py
```diff
@@ -0,0 +1,26 @@
+from transformers import pipeline
```
abrichr, line 1: Please sort imports 🙏

openadapt/tools/layout.py
```diff
@@ -0,0 +1,26 @@
...
+    else:
+        document_image = image
+
+    query_pipeline = pipeline(
+        "document-question-answering", model="impira/layoutlm-document-qa"
+    )
+
+    output = query_pipeline(document_image, question)
+    if not output:
+        logger.warning(f"{output=}")
```
abrichr, line 25:
```suggestion
        logger.warning(f"empty {output=}")
```

tests/openadapt/test_layout.py
```diff
@@ -0,0 +1,71 @@
...
+    assert output.lower() == expected_output
+
+
+def test_calc_screenshot():
+    output = document_query(
+        image=None,
+        image_path=IMAGE_FILE_NAMES[3],
+        question="What is the current number on the screen?",
+    )
+    expected_output = []
```
abrichr, line 70: Can you please confirm that we expect the model to fail to correctly answer the question here?

If that is the case, can you please suggest why this might be the case?

Ideally we would like to figure out how to get it to answer the question (e.g. by modifying the prompt). If we can't do that, we should still modify `expected_output = 10` but have pytest expect an exception here, and add a `TODO` to figure out how to correctly interpret the screenshot.

FFFiend: LayoutLM is bad at answering questions on GUI images, which is why we expect it to fail in this case. To be more specific, the pipeline just outputs an empty list as it can't parse the "10" displayed on the screen, or any number for that matter judging by its output. 

As for correctly interpreting the screenshot, that's going to be a little tricky b/c we may get an empty list as in the case of the Calculator screenshot, or some other incorrect answer making it hard to generalize.

abrichr: @FFFiend can you please implement this:

> Ideally we would like to figure out how to get it to answer the question (e.g. by modifying the prompt). If we can't do that, we should still modify expected_output = 10 but have pytest expect an exception here, and add a TODO to figure out how to correctly interpret the screenshot.

FFFiend: I'm a bit confused about the exception part. So far I've had the tool return just the list, but not an exception..so Pytest would have to expect something else

FFFiend: raise Warning for the test that's supposed to fail

tests/openadapt/test_layout.py
```diff
@@ -0,0 +1,78 @@
...
+
+def test_calendar_screenshot():
+    output = document_query(
+        image=None, image_path=IMAGE_FILE_NAMES[2], question="What month is it?"
+    )[0]["answer"]
+    expected_output = "june"
+    assert output.lower() == expected_output
+
+
+def test_calc_screenshot():
```
FFFiend, line 64: addresses the discussion here:

 https://github.com/OpenAdaptAI/OpenAdapt/pull/341#discussion_r1272928647

FFFiend:
All comments pertaining to ```test_calc_screenshot()``` have been addressed.

tests/openadapt/test_layout.py
```diff
@@ -0,0 +1,82 @@
...
+def test_calc_screenshot() -> None:
+    """Calc Screenshot test."""
+    output = document_query(
+        image=None,
+        image_path=IMAGE_FILE_NAMES[3],
+        question="What is the current number on the screen?",
+    )
+    expected_output = EXPECTED_CALC_OUTPUT
+    # Raising a warning and returning the output.
+    pytest.PytestWarning(
```
abrichr, line 76: @Mustaballer can you please confirm that this test runs and raises a warning in github, but does not cause the build to fail?

Mustaballer: I didn't see the warning being raised in the github actions log. However this is being blocked by https://github.com/OpenAdaptAI/OpenAdapt/pull/488

FFFiend: addressed in latest commit, you can check the logs to verify:
```
   /Users/runner/work/OpenAdapt/OpenAdapt/OpenAdapt/tests/openadapt/test_layout.py:76: UserWarning: This test fails because the document type is not supported.
```

abrichr:

PR #366, opened 2023-07-09
fix(utils): prevent duplicate log messages (#339) 

### LINKED ISSUE #339: [Bug]: logger produces duplicates on MacOS
### Describe the bug

During recording:

```
2023-07-02 11:33:00.922 | INFO     | __main__:read_screen_events:402 - starting                                                                                                                                        
2023-07-02 11:33:00.931 | INFO     | __main__:read_window_events:429 - starting                                                                                                                                        
2023-07-02 11:33:00.948 | INFO     | __main__:process_events:132 - starting                                                                                                                                            
2023-07-02 11:33:00.948 | INFO     | __main__:process_events:132 - starting                                
2023-07-02 11:33:01.147 | INFO     | __main__:read_window_events:448 - _window_data={'title': 'Terminal OpenAdapt — tmux — 215×51', 'left': -3, 'top': 38, 'width': 1515, 'height': 930, 'window_id': 204051}          
2023-07-02 11:33:01.147 | INFO     | __main__:read_window_events:448 - _window_data={'title': 'Terminal OpenAdapt — tmux — 215×51', 'left': -3, 'top': 38, 'width': 1515, 'height': 930, 'window_id': 204051}          
2023-07-02 11:33:02.789 | INFO     | __mp_main__:write_events:273 - event_type='screen' starting                                                                                                                       
2023-07-02 11:33:02.808 | INFO     | __mp_main__:write_events:273 - event_type='window' starting   
```

This does not occur if `utils.configure_logging` is modified to `return` immediately.

### To Reproduce

`python -m openadapt.record foo`

nweston:
Fixes a race condition in `utils.configure_logging` which leads to duplicate log messages.

**Summary**
Fixes #339.

The duplicate messages are caused by multiple threads calling `configure_logging` concurrently. This function had a race condition where two threads could both call `logger.remove()` and then both call `logger.add()`, creating two identical sinks. 

Adding a lock prevents this bug although I don't think it makes `configure_logging` entirely thread-safe. In particular, if thread A tries to log something after thread B has called `logger.remove()`, but before it calls `logger.add()`, the message might get dropped (though I haven't observed this in practice). 

I also removed some redundant `configure_logging` calls in `record.py`. This isn't necessary to fix the bug, but it seemed like a worthwhile cleanup. It's in a separate commit so it can easily be dropped if not desired.

**Checklist**
* [X] My code follows the style guidelines of OpenAdapt
* [X] I have performed a self-review of my code
* [X] If applicable, I have added tests to prove my fix is functional/effective
* [X] I have linted my code locally prior to submission
* [X] I have commented my code, particularly in hard-to-understand areas
* [X] New and existing unit tests pass locally with my changes

`python -m openadapt.record foo` will reproduce the original issue and validate the fix.

**Other information**
I saw your post on Hacker News and I might be interested in paid work on OpenAdapt. How can I learn more?

### DIFF
--- a/openadapt/record.py
+++ b/openadapt/record.py
@@ -134,7 +134,6 @@ def process_events(
         terminate_event: An event to signal the termination of the process.
     """
 
-    utils.configure_logging(logger, LOG_LEVEL)
     utils.set_start_time(recording_timestamp)
     logger.info(f"starting")
 
@@ -278,7 +277,6 @@ def write_events(
             the number of events left to be written.
     """
 
-    utils.configure_logging(logger, LOG_LEVEL)
     utils.set_start_time(recording_timestamp)
     logger.info(f"{event_type=} starting")
     signal.signal(signal.SIGINT, signal.SIG_IGN)
@@ -436,7 +434,6 @@ def read_screen_events(
         recording_timestamp: The timestamp of the recording.
     """
 
-    utils.configure_logging(logger, LOG_LEVEL)
     utils.set_start_time(recording_timestamp)
     logger.info(f"starting")
     while not terminate_event.is_set():
@@ -463,7 +460,6 @@ def read_window_events(
         recording_timestamp: The timestamp of the recording.
     """
 
-    utils.configure_logging(logger, LOG_LEVEL)
     utils.set_start_time(recording_timestamp)
     logger.info(f"starting")
     prev_window_data = {}
@@ -513,7 +509,6 @@ def performance_stats_writer(
         terminate_event: An event to signal the termination of the process.
     """
 
-    utils.configure_logging(logger, LOG_LEVEL)
     utils.set_start_time(recording_timestamp)
     logger.info("performance stats writer starting")
     signal.signal(signal.SIGINT, signal.SIG_IGN)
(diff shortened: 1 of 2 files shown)


0dm:
Thanks for the contribution!

One small thing:

looks like the logger's colours are lost 🥲

image

vs main

image

nweston:
Oops, I took the colors out for ease of grepping while I was debugging, and forgot to put them back. It's fixed now.

I used a fixup commit -- let me know if I should squash that before merging.

0dm (approved):
lgtm!

openadapt/record.py
```diff
@@ -134,7 +134,6 @@ def process_events(
         terminate_event: An event to signal the termination of the process.
     """
 
-    utils.configure_logging(logger, LOG_LEVEL)
```
abrichr, line 137: Thanks @nweston !

Can you please clarify why it was necessary to remove these calls to `utils.configure_logging`?

nweston: It's not strictly necessary -- adding the lock is sufficient to fix the bug. But these calls appear to be redundant (`utils.configure_logging` is already called at module level with the same arguments), and it's possible that there could be other race conditions, e.g. if one thread is outputting log messages while another is configuring. So removing them seemed like the safest option.

But if there's a reason to keep them, we can do that and the duplicate log messages will still be fixed.

abrichr:
Thanks @nweston ! Please grab some time on my calendar: https://www.getclockwise.com/c/richard-abrich/quick-meeting

Issue #348, opened 2023-07-04
Store screenshot diffs in database


abrichr:
### Feature request

We would like to store screenshot diffs in order to improve performance

### Motivation

https://github.com/OpenAdaptAI/OpenAdapt/blob/main/openadapt/crud.py#L232

abrichr:
Closing as part of architecture transition. Legacy frozen at v0.46.0. See PR #960.

PR #367, opened 2023-07-09
feat(crud): compute and save screenshot diff

LaPetiteSouris:

* Add 2 columns in screenshot table to store png_diff_data and png_mask_diff_data.

* CRUD now supports calculation and save screenshots diff data on the flight.

**Summary**

This PR solves https://github.com/OpenAdaptAI/OpenAdapt/issues/348. The main logic behinds the change is to introduce the function to calculate and then eventually store `png_data` difference between 2 linked screenshots. Motivation for the design:

- 2 columns that store `png_diff_data` and `png_mask_diff_data` is BLOB/binary to conform with existing `png_data`.
- However, the `Screenshot.diff` and `Screenshot.diff_mask` are kept as `PIL Image` for convenience of usage.

Side effects:
- If called, the calculation of screenshot diff is resource intensive for first time calculation, as a pairs of linked screenshot are compared, then converted to binary and store in database.
- The size of the storage significantly increases. As of now, `png_data` is the column that consumes the most space.
- It will makes space-storage worst, which is already limited by the fact that the max length of recording is ~1 minute for now.
- However, those concerns are mitigated by the fact that `precompute_diff` flag is opt-in an can be used only for case where optimization is required.

**Checklist**
* [x] If applicable, I have added tests to prove my fix is functional/effective

To activate screen diff storage, set `SAVE_SCREENSHOT_DIFF` environment variable to `True`

```bash
export SAVE_SCREENSHOT_DIFF=True
python -m openadapt.replay DemoReplayStrategy
```

This feature can be activated by an environment variable Flag. Discussion in the PR. 

- The first run should take long time as the screenshot diffs are calculated and stored.
- The second run should be instant, as the diff are only loaded from existing database.

Diff stored as BLOB in database:

Screenshot 2023-07-09 at 17 42 07


openadapt/crud.py
```diff
@@ -1,16 +1,13 @@
...
-    Screenshot,
-    Recording,
-    WindowEvent,
-    PerformanceStat,
-    MemoryStat
-)
 from openadapt.config import STOP_SEQUENCES
+from openadapt.db import Session
+from openadapt.models import (ActionEvent, MemoryStat, PerformanceStat,
+                              Recording, Screenshot, WindowEvent)
```
LaPetiteSouris, line 11: This is linted by `isort` via `precommit-hook

openadapt/models.py
```diff
@@ -1,10 +1,11 @@
...
 
-from loguru import logger
-from pynput import keyboard
-from PIL import Image, ImageChops
+import mss.tools
 import numpy as np
 import sqlalchemy as sa
+from loguru import logger
+from PIL import Image, ImageChops
+from pynput import keyboard
```
LaPetiteSouris, line 9: Changes made by `isort` in `precommit-hook`

openadapt/crud.py
```diff
@@ -221,6 +218,27 @@ def filter_stop_sequences(action_events):
...
+    if data_updated:
+        logger.info("saving screenshot diff data to db...")
+        db.bulk_save_objects(screenshots)
+        db.commit()
+
+    return screenshots
+
 
 def get_screenshots(recording, precompute_diffs=False):
     screenshots = _get(Screenshot, recording.timestamp)
```
LaPetiteSouris, line 244: This feature should be activated via an environment variable flag for convenience. By default it should be deactivated to save CPU resources.

WDYT ?

LaPetiteSouris: @0dm This is implemented now in https://github.com/OpenAdaptAI/OpenAdapt/pull/367/commits/cb14ec36c4b60afa8dfda412117f36b77a1761b5

LaPetiteSouris: Also, I just updated the PR description to facilitate manual testing the feature.

LaPetiteSouris: @0dm Any updates on my requests ?

openadapt/crud.py
abrichr: @Mustaballer can we please modify the default `isort` config to format this as `black` would (i.e. how it was before)? If not, can we add  a pre-commit hook to run black as well?

abrichr:
Thanks @LaPetiteSouris !

Please grab some time on my calendar: https://www.getclockwise.com/c/richard-abrich/quick-meeting

openadapt/crud.py
```diff
@@ -221,19 +214,37 @@ def filter_stop_sequences(action_events):
...
     for prev, cur in zip(screenshots, screenshots[1:]):
         cur.prev = prev
     screenshots[0].prev = screenshots[0]
 
-    # TODO: store diffs
-    if precompute_diffs:
-        logger.info(f"precomputing diffs...")
-        [(screenshot.diff, screenshot.diff_mask) for screenshot in screenshots]
-
+    if config.SAVE_SCREENSHOT_DIFF:
```
abrichr, line 246: @LaPetiteSouris What do you think about automatically precomputing diffs if they are not already pre-computed?

LaPetiteSouris: Doing this would ensure that we have all necessary data when we need advanced training. However, as I mentioned, it would incur heavy computing/CPU resources on first time running.

Ideally, I think we can provide 2 tracks:
- One is "basic/fast" track, which allow user to bypass all of this and have a naive but really fast results.
- One "advanced" track which incur heavy computation but provides best results.

This approach, from product side is well-round IMHO. However, as we have not figured out what and how to provide 2 experiences, I still suggest to make this computation opt-in.

WDYT ?

abrichr: Sounds good!

LaPetiteSouris: Thanks. Is there anything else you would request on this PR before an `LGTM` ? 🥇

openadapt/crud.py
```diff
@@ -321,31 +321,40 @@ def filter_stop_sequences(action_events: list[ActionEvent]) -> None:
             action_events.pop()
 
 
-def get_screenshots(
-    recording: Recording, precompute_diffs: bool = False
-) -> list[Screenshot]:
-    """Get screenshots for a given recording.
+def save_screenshot_diff(screenshots):
```
dianzrong, line 324: Please add a docstring and type annotations of the parameter and return value

openadapt/crud.py
```diff
@@ -321,31 +321,40 @@ def filter_stop_sequences(action_events: list[ActionEvent]) -> None:
...
-        recording (Recording): The recording object.
-        precompute_diffs (bool, optional): Whether to precompute screenshot diffs.
-            Defaults to False.
+    return screenshots
 
-    Returns:
-        list[Screenshot]: A list of screenshots for the recording.
-    """
+
+def get_screenshots(recording):
```
Mustaballer, line 348: Please add doctstring and type annotations for the arguments and for the return.

openadapt/crud.py
0dm: oops, sorry didn't see. lgtm

openadapt/models.py
```diff
@@ -270,24 +272,25 @@ def image(self) -> Image:
...
+                self._image = self.convert_binary_to_png(self.png_data)
         return self._image
 
     @property
-    def diff(self) -> Image:
-        """Get the difference between the current and previous screenshot."""
-        if not self._diff:
-            assert self.prev, "Attempted to compute diff before setting prev"
-            self._diff = ImageChops.difference(self.image, self.prev.image)
+    def diff(self):
```
0dm, line 279: missing type annotations

LaPetiteSouris: thanks. It's fixed

openadapt/crud.py
LaPetiteSouris: thanks. Fixed.

openadapt/crud.py
LaPetiteSouris: thanks. Fixed

PR #375, opened 2023-07-11
feat: scrub toggle for gui

### LINKED ISSUE #218: Hook up scrub toggle in GUI

0dm:
This PR closes #218 by adding a scrub toggle to the app's settings.

image

When changing a switch's value (dark / scrub), the function `config.set_env` is called to update the `.env` file. The switch values should always be synchronized with the configuration. 

There is an issue though, we can't visualize more than once in the app, so the user is told that they may need to restart the app for changes to take effect. 

```
  File "/Users/aaron/Documents/GitHub/OpenAdapt/openadapt/utils.py", line 318, in display_event
    if diff and screenshot.diff:
  File "/Users/aaron/Documents/GitHub/OpenAdapt/openadapt/models.py", line 252, in diff
    assert self.prev, "Attempted to compute diff before setting prev"
AssertionError: Attempted to compute diff before setting prev
```


openadapt/config.py
```diff
@@ -37,12 +37,14 @@
     "ACTION_TEXT_SEP": "-",
     "ACTION_TEXT_NAME_PREFIX": "<",
     "ACTION_TEXT_NAME_SUFFIX": ">",
+    # APP CONFIGURATIONS
+    "DARK_MODE": True,
```
abrichr, line 41: How about prefixing with `APP_`

openadapt/config.py
```diff
@@ -110,6 +112,28 @@ def getenv_fallback(var_name):
     return rval
 
 
+def set_env(var_name, val):
```
abrichr, line 115: Please document

openadapt/config.py
```diff
@@ -110,6 +112,28 @@ def getenv_fallback(var_name):
...
+    if not os.path.exists(".env"):
```
abrichr, line 116: Please define and refer to `ENF_FILE_PATH` instead of `.env` directly

0dm:
@abrichr comments addressed

openadapt/config.py
```diff
@@ -110,6 +114,37 @@ def getenv_fallback(var_name):
...
 
+def persist_env(var_name, val):
+    """
+    The `persist_env` function is used to persist environment variables by writing them to a `.env`
+    file.
+
+    :param var_name: The `var_name` parameter is a string that represents the name of the environment
+    variable you want to persist or update in the `.env` file
+    :param val: The `val` parameter in the `persist_env` function is the value that you want to assign
+    to the environment variable specified by `var_name`
```
abrichr, line 125: Thanks @0dm !

Can you please modify the documentation to follow the google python style guide?

Also please add `env_file_path=ENV_FILE_PATH` as a parameter.

Also please add type annotations.

0dm: [aacf116](https://github.com/OpenAdaptAI/OpenAdapt/pull/375/commits/aacf11616f9010b07dffa311045326dad3ac1415)

openadapt/app/cards.py
```diff
@@ -1,16 +1,26 @@
 import signal
```
KrishPatel13, line 1:
```
from subprocess import Popen
import signal

from nicegui import ui

from openadapt.app.objects.local_file_picker import LocalFilePicker
from openadapt.app.util import get_scrub, set_dark, set_scrub, sync_switch
```
from before import

0dm: wait.. we do it that way? I think it looks better the other way but I'll change it

0dm: [5fe7100](https://github.com/OpenAdaptAI/OpenAdapt/pull/375/commits/5fe71003e36d9aea4c5799417428662dcdf81d2b)

openadapt/config.py
```diff
@@ -110,6 +114,38 @@ def getenv_fallback(var_name):
     return rval
 
 
+def persist_env(var_name: str, val: str, env_file_path: str = ENV_FILE_PATH):
+    """
+    The `persist_env` function is used to persist environment variables by writing them to a file or
```
abrichr, line 119:
```suggestion
    Persist environment variables by writing them to a file or
```

openadapt/config.py
```diff
@@ -110,6 +114,38 @@ def getenv_fallback(var_name):
     return rval
 
 
+def persist_env(var_name: str, val: str, env_file_path: str = ENV_FILE_PATH):
+    """
+    Persist environment variables by writing them to a file or
+    updating their values if they already exist in the file.
+
+    Args:
+      var_name: The `var_name` parameter is a string that represents the name of the environment
```
abrichr, line 123:
```suggestion
      var_name: A string that represents the name of the environment
```

openadapt/config.py
```diff
@@ -110,6 +114,38 @@ def getenv_fallback(var_name):
...
 
 
+def persist_env(var_name: str, val: str, env_file_path: str = ENV_FILE_PATH):
+    """
+    Persist environment variables by writing them to a file or
+    updating their values if they already exist in the file.
+
+    Args:
+      var_name: A string that represents the name of the environment
+    variable you want to persist or update.
```
abrichr, line 124:
```suggestion
            variable you want to persist or update.
```

openadapt/config.py
```diff
@@ -110,6 +114,38 @@ def getenv_fallback(var_name):
...
 
 
+def persist_env(var_name: str, val: str, env_file_path: str = ENV_FILE_PATH):
+    """
+    Persist environment variables by writing them to a file or
+    updating their values if they already exist in the file.
+
+    Args:
+      var_name: A string that represents the name of the environment
+            variable to persist or update.
```
abrichr, line 124:
```suggestion
          variable to persist or update.
```

openadapt/config.py
```diff
@@ -110,6 +114,38 @@ def getenv_fallback(var_name):
...
+def persist_env(var_name: str, val: str, env_file_path: str = ENV_FILE_PATH):
+    """
+    Persist environment variables by writing them to a file or
+    updating their values if they already exist in the file.
+
+    Args:
+      var_name: A string that represents the name of the environment
+          variable to persist or update.
+      val: The value to assign to the environment variable.
+    environment variable specified by `var_name`.
```
abrichr, line 126:
```suggestion
```

openadapt/config.py
```diff
@@ -110,6 +114,37 @@ def getenv_fallback(var_name):
...
+      val: The value to assign to the environment variable.
+    """
+    if not os.path.exists(env_file_path):
+        with open(env_file_path, "w") as f:
+            f.write(f"{var_name}={val}")
+    else:
+        # find and replace
+        with open(".env", "r") as f:
+            lines = f.readlines()
+        for i, line in enumerate(lines):
```
abrichr, line 134: How about:

lines = [
    line
    for line in lines
    if not line.startswith(var_name)
    else f"{var_name}={val}"
]

openadapt/config.py
```diff
@@ -110,6 +114,37 @@ def getenv_fallback(var_name):
...
+            if line.startswith(f"{var_name}="):
+                lines[i] = f"{var_name}={val}\n"
+                break
```
abrichr, line 137:
```suggestion
        lines = open(ENV_FILE_PATH, "r").readlines()
        lines = [
            line
            for line in lines
            if not line.startswith(f"{var_name}=")
            else f"{var_name}={val}\n"
        ]
```

openadapt/config.py
```diff
@@ -110,6 +114,37 @@ def getenv_fallback(var_name):
...
+        else:
+            # case where "\n" is missing at end of file
```
abrichr, line 139: Why is this in an `else` block? Isn't this orthogonal to the behavior implemented by this function? What do you think about removing it?

openadapt/config.py
```diff
@@ -110,6 +114,37 @@ def getenv_fallback(var_name):
...
+            if lines[-1][-1] != "\n":
+                lines.append(f"\n{var_name}={val}")
+            else:
+                lines.append(f"{var_name}={val}")
+        with open(".env", "w") as f:
```
abrichr, line 144: `.env` should be `ENV_FILE_PATH`

openadapt/config.py
0dm: the else runs if we didn't break (didn't find the variable in .env), so it checks if there's a newline that needs to be inserted before adding the env variable

openadapt/config.py
0dm: while this logic does update the variable, it doesn't append at the end if the variable is not found. I think my for/else works better

Issue #262, opened 2023-06-14
Contrib: logo generator


abrichr:
### Feature request

Input: text description of logo components

Steps:
1. Use named entity recognition to extract entities from text description (e.g. Spacy)
2. Download SVGs from https://thenounproject.com/
3. Ask large language model to arrange
4. Generate combined SVG
5. Run through e.g. Stable Diffusion and/or ask large language model to generate animation code

Output: stylized logo image and/or animation code

### Motivation

https://github.com/OpenAdaptAI/web/issues/17

abrichr:
Related: https://arxiv.org/abs/2306.03601

abrichr:
Closing as part of architecture transition. Legacy frozen at v0.46.0. See PR #960.

PR #378, opened 2023-07-11
Contrib/logogenerator

Jonathan-Sendsen:
This PR implements a logo generator as per https://github.com/OpenAdaptAI/OpenAdapt/issues/262


contrib/wordlyworks/ner.py
```diff
@@ -0,0 +1,24 @@
...
+    text_description = '''
+    Apple Inc. is planning to open a new factory in Austin, Texas. 
+    The plan will reportedly cost $1 billion and create jobs for 5,000 people. 
+    Apple's CEO, Tim Cook, will oversee the new project.
+    '''
+
+    document_object = nlp(text_description)
+
+    for entity in document_object.ents:
+        print(f"{entity.text}, ({entity.label_})")
```
abrichr, line 16: No need to print, just return the relevant entities.

Also, please move this function to `main.py` and name it `get_entities`. It should also accept a string as an argument.

contrib/wordlyworks/main.py
```diff
@@ -0,0 +1,46 @@
...
+    logger.info(f"{key=} {secret=}")
+
+
+    auth = OAuth1(key, secret)
+    endpoint = "https://api.thenounproject.com/v2/icon/1"
+
+    response = requests.get(endpoint, auth=auth)
+    logger.info(f"response=\n{pformat(response.json())}")
+
+    if 0:
```
abrichr, line 36: Remove this

contrib/wordlyworks/main.py
```diff
@@ -0,0 +1,46 @@
...
+def main():
+    key = os.environ["NOUN_PROJECT_KEY"]
+    secret = os.environ["NOUN_PROJECT_SECRET"]
+    logger.info(f"{key=} {secret=}")
+
+
+    auth = OAuth1(key, secret)
+    endpoint = "https://api.thenounproject.com/v2/icon/1"
+
+    response = requests.get(endpoint, auth=auth)
```
abrichr, line 33: Consult the noun project API to determine which URL to use in order to search, and search for each of the entities returned by `get_entities`

abrichr:
@Jonathan-Sendsen we can run StableDiffusion in the browser with https://islamov.ai/stable-diffusion-webgpu/

contrib/wordlyworks/main.py
Jonathan-Sendsen: @abrichr just a question. Why remove all this code? The `if 0 ` part I understand why. However, the `response =...` isn't that used for the endpoint to get data from the noun project based on the API key?

Additionally, `logger.info...` I think I understand why that needs to be removed since this gets the request and prints it to the logs (for debugging purposes)

contrib/wordlyworks/main.py
Jonathan-Sendsen: Sounds good 👍  based on my current research I think we should use their `blacklist` endpoint. It allows for more control over their icons relatively compared to their other APIs.

The blacklist API is pretty neat if there are irrelevant or inappropriate terms. They will automatically be excluded compared to their `icons` endpoint. 

I'll have some stuff for us to chat about on Friday!

contrib/wordlyworks/main.py
Jonathan-Sendsen: Nevermind I think I know why we need to remove all of this because the blacklist function will send a get request on behalf of the app.

abrichr:

PR #389, opened 2023-07-17
chore: suppress identical warnings

Mustaballer:
**What kind of change does this PR introduce?**

Specify the maximum number of times an identical warning is logged and the number of times per second warnings are logged before they are **suppressed**.

**Summary**

Address https://github.com/OpenAdaptAI/OpenAdapt/issues/383

**Checklist**
* [x] My code follows the style guidelines of OpenAdapt
* [x] I have performed a self-review of my code
* [x] If applicable, I have added tests to prove my fix is functional/effective
* [x] I have linted my code locally prior to submission
* [x] I have commented my code, particularly in hard-to-understand areas
* [x] I have made corresponding changes to the documentation (e.g. README.md, requirements.txt)
* [x] New and existing unit tests pass locally with my changes

**How can your code be run and tested?**

You can specify `max_count` for how many identical warnings get logged when using openadapt library.

PR #406, opened 2023-07-18
feat: enhanced visualize.py with NiceGUI

### LINKED ISSUE #73: Refactor visualize.py with NiceGUI
https://github.com/OpenAdaptAI/OpenAdapt/blob/main/openadapt/visualize.py should be rewritten, maybe with https://nicegui.io/ , maybe something else.

We would also like to display performance information (see https://github.com/MLDSAI/OpenAdapt/issues/85)

0dm:
new visualization:

image
image
(buttons to view bigger / save plot)

now says if scrubbing:
image

events:
image
searching:
image
image

environment variables:
```
    "VISUALIZE_DARK_MODE": False,
    "RUN_NATIVELY": False,
    "DENSE_TREES": True,
    "NO_ANIMATIONS": False,
    "EXPAND_ALL": False,
    "MAX_TABLE_CHILDREN": 10,
```
outside, of visualize, I have edited utils.plot_performance with some new default parameters:
```py
def plot_performance(
    recording_timestamp: float = None,
    view_file: bool = True,
    save_file: bool = True,
    dark_mode: bool = False,
) -> str:
```
I have moved the old visualize.py to openadapt/deprecated. `python -m openadapt.deprecated.visualize` just 
as a fallback (nicegui breaks / compatibility)

this is a pretty big PR there's a lot of code (some of it might look a little strange..)


0dm:
image

openadapt/visualize1.py
```diff
@@ -0,0 +1,209 @@
...
+    configure_logging,
+    display_event,
+    image2utf8,
+    row2dict,
+    rows2dicts,
+    plot_performance,
+)
+
+SCRUB = config.SCRUB_ENABLED
+if SCRUB:
```
abrichr, line 21: how about just `if config.SCRUB_ENABLED`?

openadapt/visualize1.py
```diff
@@ -0,0 +1,209 @@
...
+    # too many warnings from scrubbing
+    __import__("warnings").filterwarnings("ignore", category=DeprecationWarning)
```
abrichr, line 23: I think we can remove this once https://github.com/OpenAdaptAI/OpenAdapt/pull/389 is merged right?

openadapt/visualize1.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+def create_tree(tree_dict, max_children=MAX_TABLE_CHILDREN):
+    tree_data = []
+    for key, value in tree_dict.items():
+        if value in EMPTY:
+            continue
+        node = {
+            "id": str(key)
+            + f"{(': '  + str(value)) if not isinstance(value, (dict, list)) else ''}"
```
abrichr, line 40: Seems like the f-string is unused here. Can you please re-write this to use the f-string?

openadapt/visualize1.py
0dm: possibly, I think there'd need to be another change to send those warnings to loguru first though.

openadapt/visualize1.py
```diff
@@ -0,0 +1,209 @@
...
+            else:
+                node["children"] = create_tree({i: v for i, v in enumerate(value)})
+        tree_data.append(node)
+    return tree_data
+
+
+def set_tree_props(tree):
+    tree._props["dense"] = True
+    tree._props["no-transition"] = True
+    # tree._props["default-expand-all"] = True
```
abrichr, line 59: How about making this configurable?

openadapt/visualize1.py
0dm: I declare SCRUB because it gets checked quite often in the main function.

openadapt/visualize1.py
0dm: Yeah I'll use this to function to fetch the newvariables from openadapt.config 👍 just had it like this for testing

0dm:
@abrichr this is ready for review

0dm:
note:
diff & mask are not being used --I will find a way to integrate them soon

Issue #414, opened 2023-07-20
Reward modelling


abrichr:
### Feature request

![image](https://github.com/OpenAdaptAI/OpenAdapt/assets/774615/0f37862f-9331-4fac-8609-7777722e030c)

### Motivation

Refer also to https://github.com/OpenAdaptAI/OpenAdapt/issues/393#issuecomment-1643936501

abrichr:
Possibly related: https://github.com/OpenAdaptAI/OpenAdapt/pull/234

abrichr:
Closing as part of architecture transition. Legacy frozen at v0.46.0. See PR #960.

Issue #227, opened 2023-06-06
Explore Microsoft Guidance


abrichr:
How can we take advantage of https://github.com/microsoft/guidance ?

For example, we want to constrain model outputs to generate valid actions. How can we use the guidance library to do so? What other features does the library offer that might be useful?

dianzrong:
I think the most useful aspect of the guidance library is to get the output of an LLM model in a pre-specified format with certain parameters [(example)](https://github.com/microsoft/guidance/blob/main/notebooks/guaranteeing_valid_syntax.ipynb). This way we can easily parse the output and turn it into a valid ActionEvent. 

However, there are limitations with Microsoft Guidance:
- very new
     - not much documentation
     - still changing; there were many commits last week
- takes lots of RAM (I ran out trying to run it. I have 8GB) and hard drive memory

There are also benefits to Microsoft Guidance over alternatives such as [LangChain](url) and [JSONFormer](https://github.com/1rgs/jsonformer):

- able to resolve issues with models tokening at the wrong spot [(token healing)](https://github.com/microsoft/guidance#token-healing-notebook)
- can easily access only certain parts of the response
- can partially generate answers and await further information [(example)](https://github.com/microsoft/guidance#token-healing-notebook)

abrichr:
Alternative: https://github.com/NVIDIA/NeMo-Guardrails

dianzrong:
[Useful features](https://github.com/microsoft/guidance/blob/f8c5a5f8f5a031b9576d4b920d9ea730c06ab4c0/notebooks/art_of_prompt_design/use_clear_syntax.ipynb#L348):

- Incremental generation which can be faster than regular generation 
- Can force the output to contain unique points
- Can use regex to specify output form
- Can create hidden blocks to stagger when the model receives additional context

abrichr:
@dianzrong you linked to TokenHealing in your link here, was that intentional?

> can partially generate answers and await further information [(example)](https://github.com/microsoft/guidance#token-healing-notebook)

abrichr:
> Incremental generation which can be faster than regular generation
> Can force the output to contain unique points
> Can use regex to specify output form
> Can create hidden blocks to stagger when the model receives additional context

This all seems extremely useful. Any interest in implementing this? 😄

dianzrong:
@abrichr I believe https://github.com/NVIDIA/NeMo-Guardrails is for those who are making LLM-powered chatbots, rather than harnessing LLMs like we are.

[...]

Issue #26, opened 2023-04-19
Implement Guardrails


abrichr:
How can we extend DemoReplayStrategy to use https://github.com/ShreyaR/guardrails ?

abrichr:
Closing as part of architecture transition. Legacy frozen at v0.46.0. See PR #960.

PR #420, opened 2023-07-23
WIP feat(model_evaluation): Add script to evaluate models

LaPetiteSouris:

Solve https://github.com/OpenAdaptAI/OpenAdapt/issues/421

**Summary**

- Build a small dataset from the last recording. The dataset contains the following attribute
`reference_window_dict`, `reference_action_dicts`, `active_window_dict` (which is also `reference_window_dict` for training dataset

- Build a simple algorithm to score a prediction. The algorithm is as simple as following

1. If the reference action and the predicted action is of different type, return 0, as they are not similar

2. If the reference action and predicted action  is of same type `press` or `release`, compare the pressed/released key. If the same key is pressed or released, return score 1, else return score 0
3.  If the reference action and predicted action is of type mouse movement, calculate the Euclidean distance between the 2 points (reference and predicted). Normalize the distance between 0-1 based on the max size of the screen. Then invert the score. I.e: 2 identical points should have score of 0 and vice-versa.

- For each entry in the dataset, build a prompt and ask for prediction from GPT-2. Then calculate the score.

**Next Steps**

Use Reinforced Learning to improve the model and re-run the above steps to compare if the score has been improved or not

**Checklist**

```bash
python -m openadapt.models_tuning.fine_tune_models

```

**Other information**

### DIFF
--- a/openadapt/models_tuning/fine_tune_models.py
+++ b/openadapt/models_tuning/fine_tune_models.py
@@ -1,0 +1,207 @@
+from copy import deepcopy
+from pprint import pformat
+import math
+
+from loguru import logger
+import fire
+import transformers as tf
+
+from openadapt import crud, models, utils
+
+LOG_LEVEL = "INFO"
+MAX_SCREEN_SIZE = (1920, 1080)
+MAX_INPUT_SIZE = 1024
+MAX_TOKENS = 1024
+
+
+@logger.catch
+def tune_model(model_name: str = "gpt2"):
+    utils.configure_logging(logger, LOG_LEVEL)
+
+    tokenizer = tf.AutoTokenizer.from_pretrained(model_name)
+    model = tf.AutoModelForCausalLM.from_pretrained(model_name)
+
+    recording = crud.get_latest_recording()
+    dataset = convert_recording_to_dataset(recording)
+
+    prediction_dataset = _generate_prediction_dataset(model, tokenizer, dataset)
+
+    ave_similarity_score = calculate_similarity_per_dataset(dataset, prediction_dataset)
+    logger.info(f"Average similarity score is {ave_similarity_score=}")
+
+
+def calculate_similarity_per_dataset(dataset, prediction):
+    total_score = 0
+    # loop through dataset and prediction pair wise
+    for entry, prediction in zip(dataset[20::], prediction[20::]):
+        reference_action_dicts = entry["reference_action_dicts"]
+        prediction_action_dicts = prediction["active_action_dicts"]
+        score_per_action = 0
+        for ref_action, prediction_action in zip(
+            reference_action_dicts, prediction_action_dicts
+        ):
+            score_per_action += _calculate_similarity_per_action(
+                ref_action, prediction_action
+            )
+        average_score = score_per_action / len(reference_action_dicts)
+        total_score += average_score
+    average_score = total_score / len(dataset)
+    return average_score
+
+
+def convert_recording_to_dataset(recording):
+    """
+    Convert a recording to a dataset.
+    """
+    dataset = []
+    for action_event in recording.processed_action_events:
+        reference_window = action_event.window_event
+        reference_window_dict = deepcopy(
+            {
+                key: val
+                for key, val in utils.row2dict(reference_window, follow=False).items()
+                if val is not None
+                and not key.endswith("timestamp")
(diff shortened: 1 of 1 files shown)


openadapt/models_tuning/fine_tune_models.py
```diff
@@ -0,0 +1,207 @@
...
+            max_length=1000,
+            num_return_sequences=1,
+        )
+        N = input_tokens["input_ids"].shape[-1]
+        completion = tokenizer.decode(
+            output_tokens[:, N:][0],
+            clean_up_tokenization_spaces=True,
+        )
+        active_action_dicts = get_action_dict_from_completion(completion)
+        logger.debug(f"active_action_dicts=\n{pformat(active_action_dicts)}")
```
LaPetiteSouris, line 184: @abrichr May be am I wrong here, but using local `gpt-2` model from `transformers` package, I do not get a lot of correct prediction. Must of the prediction given by `gpt-2` model cannot be parsed to correct `action_event`.

Is there any special tricks that I miss ?

openadapt/models_tuning/fine_tune_models.py
```diff
@@ -0,0 +1,207 @@
...
+    if len(point1) != 2 or len(point2) != 2:
+        raise ValueError("Both points must be 2D coordinates (x, y)")
+
+    x1, y1 = point1
+    x2, y2 = point2
+    distance = math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2)
+    return distance
+
+
+def _calculate_similarity_per_action(ref_action, prediction_action):
```
LaPetiteSouris, line 119: The simple algorithm is to 
- Verify if 2 actions are of the same type
- If they are both button pressing or releasing, the compare the key
- If they are mouse movement, then calculate the Euclidean between the 2 clicked point.

LaPetiteSouris: The whole point is the compare how similar the reference action to the predicted action.

openadapt/models_tuning/fine_tune_models.py
LaPetiteSouris: If this is truly the case, then either my understand of how to build the `prompt` from Event is incorrect, or we need to re-consider the way a prompt is built.

Otherwise, the whole point of improving the model becomes less critical as may be the issue is not the model, but our prompt building itself.

WDYT ? @abrichr

openadapt/models_tuning/fine_tune_models.py
```diff
@@ -0,0 +1,207 @@
...
+import math
+
+from loguru import logger
+import fire
+import transformers as tf
+
+from openadapt import crud, models, utils
+
+LOG_LEVEL = "INFO"
+MAX_SCREEN_SIZE = (1920, 1080)
```
abrichr, line 12: Can get this from @openadapt.models.recording https://github.com/OpenAdaptAI/OpenAdapt/blob/main/openadapt/models.py#L29 ?

openadapt/models_tuning/fine_tune_models.py
abrichr: @LaPetiteSouris that you for the information regarding GPT-2! We have been testing with GPT3.5-turbo and GPT-4.

For enforcing output structure, we have looked at implementing [guardrails](https://github.com/OpenAdaptAI/OpenAdapt/issues/26) and [LMQL/Guidance](https://github.com/OpenAdaptAI/OpenAdapt/issues/227), but so far without much luck. If you could assist in this as well that would be greatly appreciate 🙏 

You are correct that we need to implement evaluation metrics. Related: https://github.com/OpenAdaptAI/OpenAdapt/issues/173 https://github.com/OpenAdaptAI/OpenAdapt/issues/414

LaPetiteSouris: @abrichr I'll take a look into LMQL Guidance/guardrails topic as soon as I wrap up my work on these model evaluations/tuning topic.

openadapt/models_tuning/fine_tune_models.py
LaPetiteSouris: Thanks. I will take that into account.

LaPetiteSouris:
Close and superseded by https://github.com/OpenAdaptAI/OpenAdapt/pull/444

Issue #104, opened 2023-05-07
Implement user guidance


abrichr:
When the replay strategy determines that the current Screenshot/WindowState is sufficiently different from the expected state in the recording, we would like to prompt the user to take over and demonstrate how to complete the task in that situation:

1. compare the current state to the expected state in the current recording.
2. if not found, compare the current state to states in other recordings with the same description.
3. if not found, pause replay
4. notify the user that they need to demonstrate how to continue
5. start a new recording (with the same description as the one being recorded) as soon as the user starts generating InputEvents
6. wait for the user to finish the demonstration (e.g. by clicking on the system tray icon)
7. restart replay from where we left off

Mustaballer:
Here are some questions and thoughts I had to address for some of these steps:

- We can compare the screenshot captured during the replay to the screenshot captured during the recording(and subsequent screenshots) using computer vision tool such as OpenCV. The issue is how can we determine if its sufficiently different. 
  - Image comparison techniques such as thresholding and matching [1][2][3]. Thresholding involves setting a threshold value to determine if two images are similar or different, while matching involves comparing the features of two images using a distance calculation.
  - Is this what you had in mind?

- For the case when the user needs to demonstrate how to continue, we can use a desktop notification tool such as the WinToast library for Windows or the pynotify library for Linux.
![image](https://github.com/MLDSAI/OpenAdapt/assets/43456930/86efc4f7-8bd2-4671-89c7-de85539383fc)

- For waiting for the user to finish the demonstration (e.g. 10s of inactivity), we can use a Python timer such as the time module.

abrichr:
Also need to store this in the database, along with replays

abrichr:
Closing as part of architecture transition. Legacy frozen at v0.46.0. See PR #960.

PR #472, opened 2023-08-18
feat(replay): add user guidance

Mustaballer:

This PR introduces a feature enhancement to address [OpenAdapt Issue #104](https://github.com/OpenAdaptAI/OpenAdapt/issues/104), providing functionality for handling user interactions during replay strategies.

**Summary**

This PR addresses the need to prompt users during replay strategies when the current Screenshot/WindowState deviates from the expected state in the recording. It implements the following steps:

1. Compare the current state to the expected state in the current recording.
2. If not found, compare the current state to states in other recordings with the same description.
3. If not found, pause the replay.
4. Notify the user that they need to demonstrate how to continue.
5. Start a new recording (with the same description as the one being recorded) as soon as the user starts generating InputEvents.
6. Wait for the user to finish the demonstration (e.g., by clicking on the system tray icon).
7. Restart the replay from where we left off.

**Checklist**
* [ ] My code follows the style guidelines of OpenAdapt.
* [ ] I have performed a self-review of my code.
* [ ] If applicable, I have added tests to prove my fix is functional/effective.
* [ ] I have linted my code locally prior to submission.
* [ ] I have commented my code, particularly in hard-to-understand areas.
* [ ] I have made corresponding changes to the documentation (e.g. README.md, requirements.txt).
* [ ] New and existing unit tests pass locally with my changes.

n/a yet

**Other information**


openadapt/utils.py
```diff
@@ -24,7 +24,7 @@
 
 from openadapt import common, config
 from openadapt.db import BaseModel
-from openadapt.logging import filter_log_messages
+from openadapt.logging_config import filter_log_messages
```
Mustaballer, line 27: I was facing this error
```terminal
PSYCHOPY Error: AttributeError: module 'logging' has no attribute 'getLogger'
```

Used this solution from stackoverflow [post](https://stackoverflow.com/questions/55174754/psychopy-error-attributeerror-module-logging-has-no-attribute-getlogger):
>If the current file you are working has name logging.py change it else it will give circular import error.

Mustaballer:
@abrichr Currently I created the base code for running pyside6 application which prompts the user to to record(essentially just the UI), but have not implemented the functionality for actually recording or inserting into the db yet.  I have listed below some screenshots of the pyside application. Let me know how I can improve this!
image
image

abrichr:
What do you think about removing the "Recording in progress" screen?

As a next step, we need an API somewhere that can be triggered when running `openadapt.replay` (explicitly or by the model) that will:

1. pause replay,
2. start a recording,
3. once the recording is in progress (waiting until it's fully started is important), show the notification to the user that their assistance is required,
4. stop the recording as usual (i.e. via system tray),
5. restart replay*

* There is some subtlety here that we need to explore.

abrichr:
From our 1-on-1 conversation notes:

API should implement a singleton pattern
https://python-patterns.guide/gang-of-four/singleton/

Replay.py: (rename to control.py, or control/control.py or similar)

    _pause_event = multiprocessing.Event()

    Def replay():
        # this has been implemented
        …
        strategy.run(pause_event=_pause_event)
        …

    Def pause():
        # set the pause_even

    Def unpause()
        # clear the pause event

replay/ → control/
    __init__.py
       # import contents of replay.py (control.py)
       __all__ = [replay, pause, unpause, …]


 
Have configurable pause key in config.py to pause replay
Rename to control submodule

Refactor stop_sequence_detected to stop_sequence_event (optional)



We need to pass something into each of the processes in record.py that the process sets once it has started. E.g. a started_event. If we use multiprocessing.Event, we need a separate event per process. If we use multiprocessing.Vlaue, we can have a single event that each process increments. (The latter sounds like a leaky abstraction).

# whenever a process is created in record.py, we create a multiprocessing.Event and append it to a list
# before `while not stop_sequence_detected:`, we loop over all events in the list until they are set, at which point we set a new event, e.g. started_event
Now during replay, in a function called on_assistance_required (when it is called is TBD), we start the recording (passing in the started_event to record.py), and wait until the started_event is set, after which we can display a notification to the user

We need to confirm that the memory usage of record.py does not increase when the user is idle during a recording.

https://python-patterns.guide/gang-of-four/

abrichr:

PR #563, opened 2024-02-19
Feat/video

abrichr:
This PR
- implements video recording with `ffmpeg` via `pyav`.
- fixes a performance bug affecting the usage of `mss`
- fixes some other bugs (e.g. types, `remove_invalid_keyboard_events`, `get_timestamp`)

`config.RECORD_IMAGES == True and config.RECORD_VIDEO == False` before `mss` bugfix:

image

`config.RECORD_IMAGES == True and config.RECORD_VIDEO == False` after `mss` bugfix:

image

`config.RECORD_IMAGES == False and config.RECORD_VIDEO == True` after `mss` bugfix:

image


0dm:
What are the advantages of using FFmpeg to record instead of openadapt.capture (screen-recorder-sdk on windows, AVFoundation on Mac)? I'd guess that AVFoundation is probably more light-weight.

abrichr:
@0dm good question! The idea here is ultimately to replace the `Screenshot` model by extracting frames from the recorded video. My initial thought was that this would be easier to do with `ffmpeg`, but I'm not sure.

Do you have any thoughts on how to extract high precision frames from the video recorded by `openadapt.capture`?

PR #585, opened 2024-03-06
Use AVCaptureVideoDataOutput on capture.py

### LINKED ISSUE #570: Consolidate video.py and capture.py for local hardware acceleration
### Feature request

`capture/_macos.py` uses `AVFoundation`, `capture/_windows.py ` uses `screen_recorder_sdk` which uses `MediaFoundationAPI`. These are likely to be more performant than `mss` used in `record.py` and `video.py`, but currently `capture` does not support extracting time aligned screenshots (while `video` does):

```
(openadapt-py3.10) abrichr@MacBook-Pro-4 OpenAdapt % ffprobe captures/2024-02-19-10-43-33.mov  
ffprobe version 6.1.1 Copyright (c) 2007-2023 the FFmpeg developers
  built with Apple clang version 15.0.0 (clang-1500.1.0.2.5)
  configuration: --prefix=/usr/local/Cellar/ffmpeg/6.1.1_3 --enable-shared --enable-pthreads --enable-version3 --cc=clang --host-cflags= --host-ldflags='-Wl,-ld_classic' --enable-ffplay --enable-gnutls --enable-gpl --enable-libaom --enable-libaribb24 --enable-libbluray --enable-libdav1d --enable-libharfbuzz --enable-libjxl --enable-libmp3lame --enable-libopus --enable-librav1e --enable-librist --enable-librubberband --enable-libsnappy --enable-libsrt --enable-libssh --enable-libsvtav1 --enable-libtesseract --enable-libtheora --enable-libvidstab --enable-libvmaf --enable-libvorbis --enable-libvpx --enable-libwebp --enable-libx264 --enable-libx265 --enable-libxml2 --enable-libxvid --enable-lzma --enable-libfontconfig --enable-libfreetype --enable-frei0r --enable-libass --enable-libopencore-amrnb --enable-libopencore-amrwb --enable-libopenjpeg --enable-libopenvino --enable-libspeex --enable-libsoxr --enable-libzmq --enable-libzimg --disable-libjack --disable-indev=jack --enable-videotoolbox --enable-audiotoolbox
  libavutil      58. 29.100 / 58. 29.100
  libavcodec     60. 31.102 / 60. 31.102
  libavformat    60. 16.100 / 60. 16.100
  libavdevice    60.  3.100 / 60.  3.100
  libavfilter     9. 12.100 /  9. 12.100
  libswscale      7.  5.100 /  7.  5.100
  libswresample   4. 12.100 /  4. 12.100
  libpostproc    57.  3.100 / 57.  3.100
[mov,mp4,m4a,3gp,3g2,mj2 @ 0x7fd88a704b40] moov atom not found
captures/2024-02-19-10-43-33.mov: Invalid data found when processing input
```

This issue will be complete once we have modified these files to support saving video files recorded via `openadapt.capture` from which time-aligned screenshots can be extracted. i.e. we need to modify `openadapt.capture._macos.Capture` and `openadapt.capture._windows.Capture` to supply screenshots in memory instead of file, e.g. `self.session.addOutput_(self.file_output)`.

### Motivation

Local hardware acceleration -> maximum performance

0dm:
WIP to resolve #570

### DIFF
--- a/openadapt/capture/_macos.py
+++ b/openadapt/capture/_macos.py
@@ -4,16 +4,61 @@ This is based on: https://gist.github.com/timsutton/0c6439eb6eb1621a5964
 
 usage: see bottom of file
 """
+
 from datetime import datetime
 from sys import platform
 import os
 
 from Foundation import NSURL, NSObject  # type: ignore # noqa
 from Quartz import CGMainDisplayID  # type: ignore # noqa
+import Quartz.CoreVideo  # type: ignore # noqa
+import Quartz.CoreGraphics  # type: ignore # noqa
+from AppKit import NSImage, NSBitmapImageRep  # type: ignore # noqa
 import AVFoundation as AVF  # type: ignore # noqa
 import objc  # type: ignore # noqa
+import concurrent.futures
 
 from openadapt import config
+
+
+class VideoDelegate(NSObject):
+    """Delegate for the video capture."""
+
+    def captureOutput_didOutputSampleBuffer_fromConnection_(
+        self, output, sampleBuffer, connection
+    ):
+        image_buffer = AVF.CMSampleBufferGetImageBuffer(sampleBuffer)
+        Quartz.CoreVideo.CVPixelBufferLockBaseAddress(
+            image_buffer, Quartz.CoreVideo.kCVPixelBufferLock_ReadOnly
+        )
+        width = Quartz.CoreVideo.CVPixelBufferGetWidth(image_buffer)
+        height = Quartz.CoreVideo.CVPixelBufferGetHeight(image_buffer)
+        color_space = Quartz.CoreGraphics.CGColorSpaceCreateDeviceRGB()
+        context = Quartz.CoreGraphics.CGBitmapContextCreate(
+            Quartz.CoreVideo.CVPixelBufferGetBaseAddress(image_buffer),
+            width,
+            height,
+            8,
+            Quartz.CoreVideo.CVPixelBufferGetBytesPerRow(image_buffer),
+            color_space,
+            Quartz.CoreGraphics.kCGImageAlphaNoneSkipFirst,
+        )
+        quartz_image = Quartz.CoreGraphics.CGBitmapContextCreateImage(context)
+        Quartz.CoreVideo.CVPixelBufferUnlockBaseAddress(
+            image_buffer, Quartz.CoreVideo.kCVPixelBufferLock_ReadOnly
+        )
+        image = NSImage.alloc().initWithCGImage_size_(quartz_image, (width, height))
+        bitmap_rep = NSBitmapImageRep.alloc().initWithData_(image.TIFFRepresentation())
+        if not os.path.exists(config.CAPTURE_DIR_PATH):
+            os.mkdir(config.CAPTURE_DIR_PATH)
+        self.file_url = NSURL.fileURLWithPath_(
+            os.path.join(
+                config.CAPTURE_DIR_PATH,
+                datetime.now().strftime("%Y-%m-%d-%H-%M-%S") + ".tiff",
+            )
+        )
+        file_path = self.file_url.path()
+        bitmap_rep.TIFFRepresentation().writeToFile_atomically_(file_path, True)
 
 
 class Capture:
@@ -40,8 +85,11 @@ class Capture:
         self.screen_input = AVF.AVCaptureScreenInput.alloc().initWithDisplayID_(
             self.display_id
         )
-        self.file_output = AVF.AVCaptureMovieFileOutput.alloc().init()
+        self.video_data_output = AVF.AVCaptureVideoDataOutput.alloc().init()
         self.camera_session = None  # not used if camera=False
+
+        queue = concurrent.futures.ThreadPoolExecutor(max_workers=1)
+        delegate = VideoDelegate.alloc().init()
 
         # Create an audio device input with the default audio device
         self.audio_input = AVF.AVCaptureDeviceInput.alloc().initWithDevice_error_(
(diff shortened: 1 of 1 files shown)


abrichr:
Via ChatGPT:

Given your implementation and the issue you're facing with saving images, a few areas could potentially cause the delegate method not to trigger as expected, thereby not saving the images. Let's review and suggest corrections:

### 1. Ensure Delegate Persists

One common issue with delegates in Objective-C (and by extension, when using PyObjC) is the delegate being deallocated if it's not strongly referenced. Since you're using a concurrent futures thread pool executor as the queue for the delegate callbacks, ensure that the delegate instance (`VideoDelegate`) persists for the duration of the capture session. If it's deallocated, your callback won't be called.

In your `Capture` class, you should keep a strong reference to the `VideoDelegate` instance:

```python
self.delegate = VideoDelegate.alloc().init()
```

Then use `self.delegate` when setting the sample buffer delegate:

```python
self.video_data_output.setSampleBufferDelegate_queue_(self.delegate, queue)
```

### 2. Confirming Delegate Method Signature

Ensure the method signature in your `VideoDelegate` matches what AVFoundation expects. The method signature looks correct based on your code, but it's always a good idea to double-check this against the AVFoundation documentation and ensure no recent changes might affect it.

### 3. Debugging Delegate Calls

To verify if the issue lies with the delegate not being called, add more logging inside your delegate method and at key points in your `Capture` class setup to ensure each step is executed as expected. For example:

```python
from Foundation import NSLog

class VideoDelegate(NSObject):
    def captureOutput_didOutputSampleBuffer_fromConnection_(self, output, sampleBuffer, connection):
        NSLog("Delegate method called")
        # Your existing code...
```

### 4. Thread Pool Executor

Your use of `concurrent.futures.ThreadPoolExecutor` as the queue for delegate callbacks is unconventional. Typically, an `AVCaptureVideoDataOutput` expects a dispatch queue (created with `dispatch_queue_create`) for its second argument in `setSampleBufferDelegate_queue_`. Mixing Python's thread pool executor with Objective-C's expectation might be leading to unexpected behavior.

Change your queue to use a dispatch queue:

```python
import objc
self.queue = objc.dispatch_queue_create("videoDataOutputQueue", None)
self.video_data_output.setSampleBufferDelegate_queue_(self.delegate, self.queue)
```

### 5. Verify Configuration and Permissions

- Ensure your application has the necessary permissions to capture the screen and use the camera or microphone if enabled.
- Verify that your session configuration (adding inputs and outputs) completes successfully. If there's an issue with how the session is configured, it may not start correctly, resulting in no callbacks to the delegate.

### Summary

Adjust your code to maintain a strong reference to the delegate and use a dispatch queue for the delegate callbacks. Ensure your app has the necessary permissions and add more extensive logging to help pinpoint where the issue might be occurring. If these steps don't resolve the problem, further investigation into the session's configuration and state when trying to start it would be warranted.

openadapt/capture/_macos.py
```diff
@@ -1,23 +1,66 @@
...
+        imageBuffer = CMSampleBufferGetImageBuffer(sampleBuffer)
+        ciImage = CIImage.imageWithCVImageBuffer_options_(imageBuffer, None)
+        bitmapRep = NSBitmapImageRep.alloc().initWithCIImage_(ciImage)
+        dict = {NSImageCompressionFactor: 1.0}
+        pngData = bitmapRep.representationUsingType_properties_(NSPNGFileType, dict)
+        if not os.path.exists("captures"):
+            os.makedirs("captures")
+        pngData.writeToFile_atomically_(
+            os.path.join("captures", "frame{}.png".format(datetime.now())), False
+        )
```
abrichr, line 59: What do you think about wrapping this in a condition, e.g. `if save_images:`, with `save_images=False` as a default argument to this function?

abrichr (commented):
Excellent work @0dm !

I believe the next step is to modify [`utils.take_screenshot`](https://github.com/OpenAdaptAI/OpenAdapt/blob/main/openadapt/utils.py#L640) to return the latest version of this data.

What do you think about accepting a write queue as a parameter to `VideoDelegate`? Then that object could optionally be passed into `utils.take_screenshot`.

abrichr:
@0dm can you please remind me what is the correct way to run this?

```
(openadapt-py3.10) abrichr@MacBook-Pro-4 OpenAdapt % python openadapt/capture/_macos.py  
Press enter to stopobjc[7309]: class `NSKVONotifying_AVCaptureConnection' not linked into application

```

openadapt/capture/_macos.py
```diff
@@ -1,23 +1,66 @@
...
 
 
 class Capture:
-    """Capture the screen, audio, and camera on macOS."""
+    """Capture the screen on macOS."""
```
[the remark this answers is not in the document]
0dm: There isn't much of a use for those because we aren't capturing [movies](https://developer.apple.com/documentation/avfoundation/avcapturemoviefileoutput) anymore. I can just duplicate the old file to keep that functionality separate (call it like movie.py) or something.

openadapt/capture/_macos.py
0dm: Yeah that seems fine, just wanted to have something to show it works

0dm:
@abrichr 

> I believe the next step is to modify [utils.take_screenshot](https://github.com/OpenAdaptAI/OpenAdapt/blob/main/openadapt/utils.py#L640) to return the latest version of this data.

> What do you think about accepting a write queue as a parameter to VideoDelegate? Then that object could optionally be passed into utils.take_screenshot.

I don't think it's very easy to work directly with the VideoDelegate object because we cannot call it with arguments, we just pass it to `video_data_output.setSampleBufferDelegate_queue_`. What we could do instead is use a global write queue or maybe a named pipe from Apple's API.

> @0dm can you please remind me what is the correct way to run this?

Running it as a python module doesn't seem to work (`python -m openadapt.capture._macos`)
```py
  File "/Users/aaron/GitHub/OpenAdapt/openadapt/capture/_macos.py", line 42, in <module>
    class VideoDelegate(NSObject):
objc.error: VideoDelegate is overriding existing Objective-C class
```
So the current method is the one you had, `python3 openadapt/capture/_macos.py`. There is that warning 
```objc[7309]: class `NSKVONotifying_AVCaptureConnection' not linked into application```, but it doesn't seem to affect the functionality.

openadapt/capture/_macos.py
0dm: [3a4deff](https://github.com/OpenAdaptAI/OpenAdapt/pull/585/commits/3a4deff0d06d52c68466a2f489d12d3f326c5051)

openadapt/capture/_macos.py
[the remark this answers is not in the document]
0dm: I added it back as a separate class, MovieCapture 
[e1682ba](https://github.com/OpenAdaptAI/OpenAdapt/pull/585/commits/e1682ba7d27dae7e76432580d9c86f1c83928c0d)

abrichr:

PR #650, opened 2024-05-02
Fix tray UX

abrichr:
This PR implements a few fixes that improves the user experience when interacting with the tray:

- saves to video instead of image by default, improving perfermance
- TODO


abrichr:
```
    "RECORD_VIDEO": true,
    "RECORD_IMAGES": false,
```
image

```
    "RECORD_VIDEO": true,
    "RECORD_IMAGES": true,
```
image

```
    "RECORD_VIDEO": false,
    "RECORD_IMAGES": true,
```
image

openadapt/record.py
```diff
@@ -1205,6 +1232,10 @@ def record(
     if terminate_recording is not None:
         terminate_recording.set()
 
+    # TODO: rename on_ready
+    # TODO: consolidate terminate_recording and on_ready
```
abrichr, line 1236: @KIRA009 can you please comment on this? Happy to jump on a call to clarify if it's easier.

PR #676, opened 2024-05-17
Database access refactor

KIRA009:
This PR differentiates the access levels of the database among the various processes

**Summary**
This PR differentiates the access levels of the database among the various processes, and introduces two conventions.

- Processes which only require data from the database, without wanting to write back to it, should use `crud.get_new_session(read_only=True)`
- Processes that require write access, must first acquire the write lock, then use the session returned by `crud.get_new_session(read_and_write=True)` to make changes to the database. Once all operations are complete, the process must release the lock using `crud.release_db_lock`. By default, any other process trying to acquire the lock will wait for 60 seconds, after which it will return `False`.
- The lock in question, is a simple txt file suffixed by `.lock`, which contains information about the process id of the process which acquired the lock, and when the lock was acquired. When the lock is released, the file is deleted, and any other process can acquire the lock by creating a new file. This will allow multiple disparate processes to share the idea of locking of database writes, given that they follow the convention described above, and is fairly safe from race conditions given that we are unlikely to face high number of write requests at the same time.
  - Ideally, at the start of the `python -m openadapt.entrypoint`, there should be no lockfile, but just in case, running the `entrypoint` module, cleans the lock file and starts anew, so special attention needs to be put here.

**Checklist**

**Other information**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -50,7 +50,7 @@ def setup_database(request: pytest.FixtureRequest) -> engine:
 
 
 @pytest.fixture(scope="session")
-def test_database(request: pytest.FixtureRequest) -> engine:
+def db_engine(request: pytest.FixtureRequest) -> engine:
     """Return the database engine for testing."""
     db_url = DATA_DIR_PATH / "test.db"
     engine = create_engine(f"sqlite:///{db_url}")
--- a/tests/openadapt/test_crud.py
+++ b/tests/openadapt/test_crud.py
@@ -10,16 +10,16 @@ from openadapt.db import crud, db
 from openadapt.models import Recording
 
 
-def test_get_new_session_read_only(test_database: sa.engine.Engine) -> None:
+def test_get_new_session_read_only(db_engine: sa.engine.Engine) -> None:
     """Test that get_new_session returns a read-only session when read_only=True.
 
     Args:
-        test_database (sa.engine.Engine): The test database engine.
+        db_engine (sa.engine.Engine): The test database engine.
     """
     # patch the ReadOnlySession class to return a mock object
     with patch(
         "openadapt.db.db.get_read_only_session_maker",
-        return_value=db.get_read_only_session_maker(test_database),
+        return_value=db.get_read_only_session_maker(db_engine),
     ):
         session = crud.get_new_session(read_only=True)
         recording = Recording(


experiments/imagesimilarity.py
```diff
@@ -290,7 +289,8 @@ def display_distance_matrix_with_images(
 
 def main() -> None:
     """Main function to process images and display similarity metrics."""
-    recording = crud.get_latest_recording()
+    db = crud.get_new_session(read_only=True)
```
abrichr, line 292: What do you think about naming this `session` for consistency?

openadapt/app/dashboard/api/recordings.py
```diff
@@ -59,40 +56,37 @@ def recording_detail_route(self) -> None:
...
-                    if event_dict["canonical_key"]:
-                        event_dict["canonical_key"] = str(event_dict["canonical_key"])
-                    if event_dict["reducer_names"]:
-                        event_dict["reducer_names"] = list(event_dict["reducer_names"])
-                    await websocket.send_json(
-                        {"type": "action_event", "value": event_dict}
-                    )
-
-                await websocket.close()
+            db = crud.get_new_session(read_only=True)
```
abrichr, line 59: As above, what do you think about naming this `session` for consistency?

(This patterns appears multiple times in this PR, I will not address each additional one.)

openadapt/app/tray.py
```diff
@@ -77,6 +76,8 @@ def __init__(self) -> None:
 
         self.app.setQuitOnLastWindowClosed(False)
 
+        crud.release_db_lock()
```
abrichr, line 79: It seems odd to release the lock before acquiring it here. What do you think about adding a comment explaining why this is necessary?

openadapt/db/db.py
```diff
@@ -76,6 +76,7 @@ def get_base(engine: sa.engine) -> sa.engine:
 engine = get_engine()
 Base = get_base(engine)
 Session = sessionmaker(bind=engine)
+ReadOnlySession = sessionmaker(bind=engine, autoflush=False, autocommit=False)
```
abrichr, line 79: This appears to not be read-only, since the user can still write if they manually call `commit()` on the session. What do you think about naming it `NonLockingSession`, and renaming the arguments to get_new_session to be `non_locking` and `standard` or similar?

openadapt/db/crud.py
```diff
@@ -420,71 +457,75 @@ def save_screenshot_diff(screenshots: list[Screenshot]) -> list[Screenshot]:
...
     Returns:
         sa.orm.Session: A new database session.
     """
+    assert not (
+        read_only and read_and_write
+    ), "Cannot be both read-only and read-and-write."
+    assert read_only or read_and_write, "Must be either read-only or read-and-write."
+    if read_only:
+        return ReadOnlySession()
     return Session()
```
abrichr, line 519: I wonder if there is a simpler approach.

Can you please evaluate each of these options suggested by ChatGPT to see if they resolve the issue without using a separate session?

---

1. **Connection Configuration**:
   - Use `isolation_level` to set the transaction behavior in SQLAlchemy. For SQLite, setting the `isolation_level` to `"IMMEDIATE"` can be useful. It starts a transaction immediately but doesn’t lock the database until a write operation is performed. This allows reads to happen concurrently until a write is needed.
   ```python
   engine = create_engine('sqlite:///yourdb.db', isolation_level="IMMEDIATE")
   NonLockingSession = sessionmaker(bind=engine, autoflush=False, autocommit=False)
   ```

2. **Database File Configuration**:
   - Ensure that your SQLite database file is on a fast storage medium to reduce the time any lock is held.
   - Consider using the `WAL` (Write-Ahead Logging) mode if your application can handle it. WAL can allow concurrent reads and writes. Enable WAL mode using a command like this after establishing a connection:
   ```python
   engine.execute("PRAGMA journal_mode=WAL;")
   ```

abrichr (changes requested):
Thanks for putting this together @KIRA009 ! I left a few comments, happy to chat about any of it if you like.

openadapt/db/db.py
abrichr: Good idea, let's do it!

abrichr: As discussed, please add a test 🙏

Issue #614, opened 2024-04-09
Avoid unnecessary segmentation + description in `VisualReplayStrategy`


abrichr:
### Feature request

https://github.com/OpenAdaptAI/OpenAdapt/pull/610 introduced the `VisualReplayStrategy` which works by segmenting the active window for every mouse event.

This is wasteful because some or all of the active window may not change between mouse events.

We would like to implement the following optimization:

1. Store the segmentation retrieved in https://github.com/OpenAdaptAI/OpenAdapt/pull/610/files#diff-4123d48b6e604812e5bbba6507183956b05038539947eedfd02a7e475344cbc5R313 (i.e. the `Segmentation` object) in the database.

2. During replay, in the `VisualReplayStrategy`, find the active window screenshot that is most similar to the current active window, e.g. using https://github.com/JohannesBuchner/imagehash. (Retrieve all `Screenshot`s for the recording, and extract the active window with https://github.com/OpenAdaptAI/OpenAdapt/blob/main/openadapt/models.py#L315.)

3. Extract the portion of the active window that is different from the  active window retrieved in step 2, and segment + describe only this portion, rather than the full thing.

### Motivation

VisualReplayStrategy is very slow.

abrichr:
- https://github.com/JohannesBuchner/imagehash
- https://github.com/facebookresearch/faiss
- https://github.com/up42/image-similarity-measures

abrichr:
Closing as part of the architecture transition to modular packages. The legacy codebase has been frozen at v0.46.0.

For the new modular architecture, please open issues in the relevant package repository:
- GUI capture: https://github.com/OpenAdaptAI/openadapt-capture
- ML/training: https://github.com/OpenAdaptAI/openadapt-ml
- Benchmarks: https://github.com/OpenAdaptAI/openadapt-evals
- Privacy: https://github.com/OpenAdaptAI/openadapt-privacy

See PR #960 for details on the new architecture.

PR #679, opened 2024-05-21
Handle similar segments

abrichr:
Related: https://github.com/OpenAdaptAI/OpenAdapt/issues/692

Work in progress

excel

image

image

image
image

nms.py:
image

Using accessibility data (only available for Excel on Windows):

image


KrishPatel13:
@abrichr , I would like to know more about the fix in this PR.

I have understood that the `VisualReplayStrategy` makes _duplicate segmentation_. Hence, we want to fix it. Correct me if I have thought incorrect.

Moreover, I need more clarification about the 2 TODOs. What do we want by sliding window and nms (does this mean non-maxima supression) ?

Thank you.

abrichr:
Thanks for your interest @KrishPatel13 !

> I have understood that the VisualReplayStrategy makes duplicate segmentation. Hence, we want to fix it. Correct me if I have thought incorrect.

This is actually a separate issue tracked in https://github.com/OpenAdaptAI/OpenAdapt/issues/614. It's more of a performance optimization.

> Moreover, I need more clarification about the 2 TODOs. What do we want by sliding window and nms (does this mean non-maxima supression) ?

There are two separate but related issues here:

1. The way that the VisualReplayStrategy works is to first have the model describe the available segments, then modify the actions based on the user's instructions and the segment descriptions. A problem arises if there are multiple visually similar segments (e.g. cells in a spreadsheet). Initially I was thinking that this could be solved by prompting the model to describe the differences between similar segments. Another more deterministic way to solve this is to include "grid landmarks" in each segment description. That is, simply include in the element description the description of other elements that are aligned with it horizontally and vertically.

2. A single run of any segmentation model will not pick up every segment. Sliding Window + Non-Maximum Suppression is a possible solution (used in https://github.com/google-research/scenic/tree/main/scenic/projects/owl_vit). I have created a new issue for this here: https://github.com/OpenAdaptAI/OpenAdapt/issues/695. I propose you take this on.

What do you think?

Issue #715, opened 2024-06-06
Add Posthog to system tray


abrichr:
### Feature request

We would like to view system tray events in posthog

### Motivation

Understanding user activity

abrichr:
Via ChatGPT:
```
import posthog
from PySide6.QtWidgets import QApplication, QSystemTrayIcon, QMenu, QAction

# Initialize PostHog
posthog.api_key = 'your-secret-api-key'
posthog.host = 'https://app.posthog.com'  # Use your PostHog instance URL here

class SystemTrayIcon:
    def __init__(self) -> None:
        self.app = QApplication([])
        self.tray = QSystemTrayIcon(QIcon("path/to/icon.png"), self.app)
        self.menu = QMenu()
        self.setup_menu()
        self.tray.setContextMenu(self.menu)
        self.tray.show()

    def setup_menu(self):
        record_action = QAction("Record", self)
        record_action.triggered.connect(self.record)
        self.menu.addAction(record_action)

        quit_action = QAction("Quit", self)
        quit_action.triggered.connect(self.quit)
        self.menu.addAction(quit_action)

    def record(self):
        # Track event
        posthog.capture('user-id', 'Record Button Clicked', {
            'timestamp': datetime.now().isoformat()
        })
        print("Recording started...")  # Your recording logic here

    def quit(self):
        # Track event
        posthog.capture('user-id', 'Quit Button Clicked', {
            'timestamp': datetime.now().isoformat()
        })
        self.app.quit()

    def run(self):
        self.app.exec_()
```

abrichr:
In addition to this  it would be good to capture the following additional function calls (regardless of where they are called from):
```
openadapt.record
openadapt.replay
openadapt.visualize
openadapt.events.get_events
```

PR #737, opened 2024-06-12
Add posthog analytics to tray actions

KIRA009:
This PR introduces posthog analytics to server actions

**Summary**
This PR adds a couple of things
- Adds a unique identifier for users (set by running migrations) to the config, so its easier to group events from different users
- Adds psothog events to server side functions and tray buttons

#715 

**Checklist**

To test locally, search for `posthog.disabled` in `openadapt/utils.py` and comment out that code block. Also in `openadapt/app/dashboard/run.py`, set `NEXT_PUBLIC_MODE` to "production" unconditionally, just to mimic what the app would behave like when run from the build.

Then run `python -m openadapt.entrypoint` (which is the usual command to run the app) and the new migration should create a unique user id in your `config.json` file. Play around the tray and the dashboard, and you should be able to see posthog events coming in tied to your user id (https://us.posthog.com/project/68185/persons)


abrichr (approved):
LGTM!

Issue #768, opened 2024-06-17
Fix Posthog events


abrichr:
![Image](https://github.com/OpenAdaptAI/OpenAdapt/assets/774615/b897e2b4-14e7-4deb-942c-87c3d1102da4)

- we would like there to be zero "None" events (all of these should have some more descriptive categorization)
- events named after a recording (e.g. "text someone") should be combined into the "delete", "replay", or "visualize" event

KIRA009:
This has been addressed

PR #772, opened 2024-06-18
Posthog event logging updates

KIRA009:
This PR adds a version number of the openadapt build to posthog events, so its easier to debug the source of erroneous events. It also adds a fix so that submenus in the tray, that have the name of the recording are not reported, rather the underlying action is reported

**Summary**
Address #768 

**Checklist**

In `utils.py`, comment out the lines that disable posthog in the `get_posthog_instance` function. I have added a property to psothog events that will differentiate dev events from events coming from the app, so this can be used when we need to test without polluting the events charts. Play around with the tray (also the submenus) and check that the events are logged correctly

### DIFF
--- a/openadapt/app/tray.py
+++ b/openadapt/app/tray.py
@@ -62,13 +62,18 @@ class TrackedQAction(QAction):
             text (str): The text of the action.
             parent (QWidget): The parent widget.
         """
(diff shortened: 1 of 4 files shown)


abrichr:
@KIRA009 can you please fix merge conflicts? I believe we should pass in `recording` to `get_perf_stats`

Issue #755, opened 2024-06-14
Implement automatic updates


abrichr:
### Feature request

We would like the OpenAdapt app to automatically update on launch.

### Motivation

https://github.com/OpenAdaptAI/OpenAdapt/issues/754

abrichr:
Closing as part of the architecture transition to modular packages. The legacy codebase has been frozen at v0.46.0.

Automatic updates were for the legacy app which has been replaced. The new modular packages are installed via pip.

For the new modular architecture, please open issues in the relevant package repository:
- GUI capture: https://github.com/OpenAdaptAI/openadapt-capture
- ML/training: https://github.com/OpenAdaptAI/openadapt-ml
- Benchmarks: https://github.com/OpenAdaptAI/openadapt-evals
- Privacy: https://github.com/OpenAdaptAI/openadapt-privacy

See PR #960 for details on the new architecture.

PR #782, opened 2024-06-19
Feat : Add download update feature in user app. Issue #755

shashank40:
/claim #755

Feature addition

**Summary**

Currently we don't support app update functionality in our app. With this addition, user can get the latest version of app from the app-tray itself.

**Checklist**

**Other information**


abrichr:
@shashank40 this a great start!

In the comment you linked, the third step is:

> On clicking that, the newest version should be downloaded, and unzipped in the same directory as where the current app lives. There are no additional installation instructions for OpenAdapt, so we could just display a message box saying that the new app was downloaded, and maybe delete the current version as well (@abrichr thoughts on this as well)

If you implement this we can consider this. Keep going!

shashank40:
@abrichr
Just a thought!
According to me unzipping in the download directory  better than current directory as people generally look for downloads in the downloads folder rather than any random directory)? (Your thoughts @abrichr?)

I think that is the only part left in the comment.

But i will implement the unzipping part for sure.

shashank40:
Also @abrichr
i dont think we should implement deleting the old version. 
**Reason :** what if the user is in a middle of the recording during download? And we delete this old build, this is not ideal

I think extracting the zip is the best thing we should do and not the deleting part.

Thoughts? @abrichr

abrichr:
@shashank40 typically how applications handle this is to say something like:

> A new version is available: vxx.xx.xx (you are on vyy.yy.yy). Would you like to update? Yes/No/Skip this version

We can show a message on startup indicating that a new version is available, and let the user decide how to proceed.

shashank40:
@abrichr 
Yess i agree, but isn't this similar to what is implemented but better?. Rather than popup we have it in a tray with the version to update visible.
User can choose to implement and not to implement.
 
 Why i say this is, as OpenAdapt gets latest updates **quite frequently**(sometimes even 2-3 a day) and not weekly/monthly. So let's say i update it once and then few hours later there would be another popup hindering me. 
 
 This would be a bad user experience every time i open an app.
 
 The implementation that you suggested might work better for apps/softwares that have scheduled updates that come weekly/monthly. But with these frequent updates, our user experience will go down.

abrichr:
It's an interesting point @shashank40. But I still think we want to make the user's life easier by automatically extracting the application for them if they do choose to update.

shashank40:
@abrichr You mean extracting the zip file?

shashank40:
@abrichr Done.
Added the functionality to unzip as well

abrichr:
@shashank40 I'm unable to test this on my machine because of the following:

```
136272 INFO: Converting EXE to target arch (arm64)                                                                                                                                                                                                                                      09:37:57 [256/80932]
136496 INFO: Removing signature(s) from EXE                                                                                                                                                                                                                                                                 
136569 INFO: Appending PKG archive to EXE                                                                                                                                                                                                                                                                   
136682 INFO: Fixing EXE headers for code signing                                                                                                                                                                                                                                                            
136684 INFO: Re-signing the EXE                                                                                                                                                                                                                                                                             
137030 INFO: Building EXE from EXE-00.toc completed successfully.                                                                                                                                                                                                                                           
137265 INFO: checking COLLECT                                                                                                                                                                                                                                                                               
137265 INFO: Building COLLECT because COLLECT-00.toc is non existent                                                                                                                                                                                                                                        
137265 INFO: Building COLLECT COLLECT-00.toc                                                                                                                                                                                                                                                                
Traceback (most recent call last):                                                                                                                                                                                                                                                                          
  File "/Users/abrichr/Library/Caches/pypoetry/virtualenvs/openadapt-VBXg4jpm-py3.10/bin/pyinstaller", line 8, in <module>                                                                                                                                                                                  
    sys.exit(_console_script_run())                                                                                                                                                                                                                                                                         
  File "/Users/abrichr/Library/Caches/pypoetry/virtualenvs/openadapt-VBXg4jpm-py3.10/lib/python3.10/site-packages/PyInstaller/__main__.py", line 228, in _console_script_run                                                                                                                                
    run()                                                                                                                                                                                                                                                                                                   
  File "/Users/abrichr/Library/Caches/pypoetry/virtualenvs/openadapt-VBXg4jpm-py3.10/lib/python3.10/site-packages/PyInstaller/__main__.py", line 212, in run                                                                                                                                                
    run_build(pyi_config, spec_file, **vars(args))                                                                                                                                                                                                                                                          
  File "/Users/abrichr/Library/Caches/pypoetry/virtualenvs/openadapt-VBXg4jpm-py3.10/lib/python3.10/site-packages/PyInstaller/__main__.py", line 69, in run_build                                                                                                                                           
    PyInstaller.building.build_main.main(pyi_config, spec_file, **kwargs)                                                                                                                                                                                                                                   
  File "/Users/abrichr/Library/Caches/pypoetry/virtualenvs/openadapt-VBXg4jpm-py3.10/lib/python3.10/site-packages/PyInstaller/building/build_main.py", line 1139, in main                                                                                                                                   
    build(specfile, distpath, workpath, clean_build)                                                                                                                                                                                                                                                        
  File "/Users/abrichr/Library/Caches/pypoetry/virtualenvs/openadapt-VBXg4jpm-py3.10/lib/python3.10/site-packages/PyInstaller/building/build_main.py", line 1079, in build                                                                                                                                  
    exec(code, spec_namespace)                                                                                                                                                                                                                                                                              
  File "OpenAdapt.spec", line 40, in <module>                                                                                                                                                                                                                                                               
    coll = COLLECT(                                                                                                                                                                                                                                                                                         
  File "/Users/abrichr/Library/Caches/pypoetry/virtualenvs/openadapt-VBXg4jpm-py3.10/lib/python3.10/site-packages/PyInstaller/building/api.py", line 1092, in __init__                                                                                                                                      
    self.__postinit__()                                                                                                                                                                                                                                                                                     
  File "/Users/abrichr/Library/Caches/pypoetry/virtualenvs/openadapt-VBXg4jpm-py3.10/lib/python3.10/site-packages/PyInstaller/building/datastruct.py", line 184, in __postinit__                                                                                                                            
    self.assemble()                                                                                                                                                                                                                                                                                         
  File "/Users/abrichr/Library/Caches/pypoetry/virtualenvs/openadapt-VBXg4jpm-py3.10/lib/python3.10/site-packages/PyInstaller/building/api.py", line 1138, in assemble                                                                                                                                      
    src_name = process_collected_binary(                                                                                                                                                                                                                                                                    
  File "/Users/abrichr/Library/Caches/pypoetry/virtualenvs/openadapt-VBXg4jpm-py3.10/lib/python3.10/site-packages/PyInstaller/building/utils.py", line 309, in process_collected_binary                                                                                                                     
    osxutils.binary_to_target_arch(cached_name, target_arch, display_name=src_name)                                                                                                                                                                                                                         
  File "/Users/abrichr/Library/Caches/pypoetry/virtualenvs/openadapt-VBXg4jpm-py3.10/lib/python3.10/site-packages/PyInstaller/utils/osx.py", line 351, in binary_to_target_arch                                                                                                                             
    raise IncompatibleBinaryArchError(                                                                                                                                                                                                                                                                      
PyInstaller.utils.osx.IncompatibleBinaryArchError: /Users/abrichr/Library/Caches/pypoetry/virtualenvs/openadapt-VBXg4jpm-py3.10/lib/python3.10/site-packages/sqlalchemy/cresultproxy.cpython-310-darwin.so is incompatible with target arch arm64 (has arch: x86_64)!                                       
Traceback (most recent call last):                                                                                                                                                                                                                                                                          
  File "/opt/homebrew/Cellar/python@3.10/3.10.14/Frameworks/Python.framework/Versions/3.10/lib/python3.10/shutil.py", line 816, in move                                                                                                                                                                     
    os.rename(src, real_dst)                                                                                                                                                                                                                                                                                
FileNotFoundError: [Errno 2] No such file or directory: '/Users/abrichr/oa/OpenAdapt/dist/OpenAdapt.app/Contents/MacOS/OpenAdapt' -> '/Users/abrichr/oa/OpenAdapt/dist/OpenAdapt.app/Contents/MacOS/OpenAdapt.app'                                                                                          
                                                                                                                                                                                                                                                                                                            
During handling of the above exception, another exception occurred:                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                            
Traceback (most recent call last):                                                                                                                                                                                                                                                                          
  File "/Users/abrichr/oa/OpenAdapt/openadapt/build.py", line 180, in <module>                                                                                                                                                                                                                              
    main()                                                                                                                                                                                                                                                                                                  
  File "/Users/abrichr/oa/OpenAdapt/openadapt/build.py", line 164, in main                                                                            
    shutil.move(                                                           
  File "/opt/homebrew/Cellar/python@3.10/3.10.14/Frameworks/Python.framework/Versions/3.10/lib/python3.10/shutil.py", line 836, in move                                                                                                                                                                     
    copy_function(src, real_dst)                                           
  File "/opt/homebrew/Cellar/python@3.10/3.10.14/Frameworks/Python.framework/Versions/3.10/lib/python3.10/shutil.py", line 434, in copy2                                                                                                                                                                    
    copyfile(src, dst, follow_symlinks=follow_symlinks)                    
  File "/opt/homebrew/Cellar/python@3.10/3.10.14/Frameworks/Python.framework/Versions/3.10/lib/python3.10/shutil.py", line 254, in copyfile                                                                                                                                                                 
    with open(src, 'rb') as fsrc:                                          
FileNotFoundError: [Errno 2] No such file or directory: '/Users/abrichr/oa/OpenAdapt/dist/OpenAdapt.app/Contents/MacOS/OpenAdapt'      
```

Can you please record a video showing the updated functionality?

openadapt/app/tray.py
```diff
@@ -168,6 +183,67 @@ def _quit() -> None:
         # for storing toasts that should be manually removed
         self.sticky_toasts = {}
 
+    def stop_download(self) -> None:
+        global CANCEL_APP_DOWNLOAD
```
abrichr, line 187: What do you think about using a property instead of globals?

abrichr:
@shashank40 if you can upload a video showing this working, and fix the formatting issues causing the build to fail, then I think we can get this merged!

shashank40:
Surely @abrichr , will update this in sometime. Currently AFK.

openadapt/build_utils.py
```diff
@@ -3,7 +3,20 @@
 import os
 import pathlib
 import sys
+import shutil
 
+def set_permissions(path):
```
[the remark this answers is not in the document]
shashank40: I would say set permission is more of a utility function. This can be used in future for any unzipping that we do.

openadapt/app/tray.py
shashank40: @abrichr  We cannot use properties here.
Reason for this is that i am using `global` to communicate between multiple threads which cannot be done using properties as the **main thread** would be different and the download will happen in a **background thread**.

Even if i set the property in a **background thread** it wont be communicated to the **main thread**. So global would be the way to communicate between them

openadapt/app/tray.py
```diff
@@ -168,6 +183,67 @@ def _quit() -> None:
         # for storing toasts that should be manually removed
         self.sticky_toasts = {}
 
+    def stop_download(self) -> None:
```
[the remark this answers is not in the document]
shashank40: According to me these functions are related to the system tray?
And wont be related to any other functionality in the system. They make more sense to be the property of system tray rather than other file.

shashank40:
Hi @abrichr.

I have replied back to the comments. I am also adding a new video here.

In the video you can see the latest version now shows as 0.34.0 and in yesterday's video it was 0.33.2 and so this makes the app quite reactive to new updates. That will be the beauty of this new update. Dynamically telling you the latest version available.

The video is mostly download, to show you the full download(you can move that part ahead). At the end it shows how new file is unzipped.

https://github.com/OpenAdaptAI/OpenAdapt/assets/48802836/c9a4cd16-44a1-4ad2-9aa9-c5a5ac8421b0

shashank40:
Hey @abrichr  can we go ahead and merge this?

shashank40:
@abrichr any updates on this?

shashank40:
@abrichr any update on this?

abrichr:
Hi @shashank40 , this looks good. However it's unfortunate that there is no visual indicator for the user about the progress, other than what is in the terminal (users who run the app not from the terminal won't see any progress).

I think we need some sort of status indicator, e.g. via `show_toast`. What do you think?

shashank40:
Yess @abrichr , I can add a toast after the zipping is complete.
Will do it in right now.

shashank40:
@abrichr Showing download complete at the end would fix the issue i suppose?

shashank40:
@abrichr As the video would get too long to upload, i am adding 2 parts of the video.
Hope this is fine?

1) Shows how the toast would popup with approximate time to download.

https://github.com/OpenAdaptAI/OpenAdapt/assets/48802836/939da167-3ec2-4d8f-a6ff-4aa293072655


2) Directly shows the end part , where after download a toast pops up with download complete

https://github.com/OpenAdaptAI/OpenAdapt/assets/48802836/a47d0708-4458-4563-a9f5-e185acf2187e

abrichr:
@shashank40 this is not bad but I think we can do better.

Can you please add a callback that clears the current toast, and updates it with a new one with an updated ETA?

shashank40:
@abrichr I initially did that, but as our main thread will be busy doing much more important things like `recording/visualize`, making main thread do regular UI upgrades(every few milliseconds) is degrading performance as main thread chokes. 

I am okay with updating the time every few seconds, but i think giving an approximate at the ends works okay and does not hinder with more important tasks

abrichr:
No need to update every few milliseconds, but I believe every 1-10s would be much more usable at minimal performance cost. 🙏

shashank40:
@abrichr Okay, so i will do a new toast pop-up every 2 secs with updated time estimate? Will that be fine?

abrichr:
At the same time be sure to clear the previous toast so that we don't ever have more than one toast showing regarding the update progress. Ideally it looks like the same toast, so remove any animations etc.

shashank40:
Yess @abrichr i made sure of that. Here is the video.
Hopefully now we can get it merged. It's been too long now 😅

https://github.com/OpenAdaptAI/OpenAdapt/assets/48802836/68819acd-b467-43a9-9f1d-3982082d3209

abrichr:
Hi @shashank40 this is looking good! I haven't merged it yet because I saw you pushed a few commits after your last comment. Please fix the merge conflicts and let me know when this is ready for re-review, would love to get this merged! 🙏

build_scripts/get_version.py
```diff
@@ -1,12 +1,25 @@
...
 def get_version() -> str:
     """Get the version of the package."""
     return importlib.metadata.version("openadapt")
 
 
+def get_latest_version() -> str:
+    """Get the latest version of the app available."""
+    response = requests.get(APP_VERSION_URL, stream=True)
+    data = json.loads(response.text)
+    return data[0]["tag_name"].replace("v", "")
```
abrichr, line 21: What do you think about extracting this from `pyproject.toml` instead? e.g.

```
def get_latest_version() -> str:
    """Get the latest version of the app available."""
    toml_path = Path(config.PARENT_DIR_PATH) / "pyproject.toml"
    parser = TomlFileParser(str(toml_path))
    parsed_data = parser.parse()
    latest_version = parsed_data['tool']['poetry']['version']
    return latest_version
```

openadapt/app/tray.py
abrichr: I think we can use a `threading.Event` here, which is more thread-safe. e.g.:
```
class SystemTrayIcon(QObject):
    def __init__(self) -> None:
        self.cancel_download_event = threading.Event()

    def stop_download(self) -> None:
        self.cancel_download_event.set()

    @Slot(str)
    def download_start_toast_slot(self, message: str) -> None:
        self.cancel_download_event.clear()

    def download_latest_version(self, base_url: str, latest_version: str) -> None:
        DOWNLOAD_URL = ""
        FILE_NAME = ""

        if sys.platform == "darwin":
            FILE_NAME = f"OpenAdapt-v{latest_version}.app.zip"
            DOWNLOAD_URL = base_url + f"/v{latest_version}/{FILE_NAME}"
        else:
            FILE_NAME = f"OpenAdapt-v{latest_version}.zip"
            DOWNLOAD_URL = base_url + f"/v{latest_version}/{FILE_NAME}"

        downloads_path = os.path.join(os.path.expanduser("~"), "Downloads")
        local_filename = os.path.join(downloads_path, FILE_NAME)

        response = requests.get(DOWNLOAD_URL, stream=True)
        total_size = response.headers.get("content-length")
        total_size = int(total_size) if total_size else None
        block_size = 1024  # 1 Kilobyte
        start_time = time.time()
        last_update_time = start_time
        with open(local_filename, 'wb') as f:
            for data in response.iter_content(block_size):
                if self.cancel_download_event.is_set():
                    break
                f.write(data)
```

build_scripts/get_version.py
shashank40: No @abrichr.
That only works in local development as we take a gitpull and pyproject.toml would get updated. 
But in case of users that won't happen and this github version would be our source of truth.

openadapt/app/tray.py
shashank40: Yess @abrichr this could be done in place of `global`
I have made the necessary change and it now uses `threading.Event()`.

shashank40:
@abrichr I did all the required changes. Removed the conflict.
Hopefully this now looks all good.

shashank40:
@abrichr Any update?

openadapt/app/tray.py
```diff
@@ -185,6 +218,108 @@ def _quit() -> None:
...
+        """Download latest version of the app."""
+        DOWNLOAD_URL = ""
+        FILE_NAME = ""
+
+        if sys.platform == "darwin":
+            FILE_NAME = f"OpenAdapt-v{latest_version}.app.zip"
+            DOWNLOAD_URL = base_url + f"/v{latest_version}/{FILE_NAME}"
+        else:
+            FILE_NAME = f"OpenAdapt-v{latest_version}.zip"
+            DOWNLOAD_URL = base_url + f"/v{latest_version}/{FILE_NAME}"
```
abrichr, line 256:
```suggestion
        if sys.platform == "darwin":
            file_ext = ".app.zip"
        else:
            file_ext = ".zip"

        file_name = f"open_adapt-v{latest_version}{file_ext}"
        download_url = base_url + f"/v{latest_version}/{file_name}"
```

openadapt/app/tray.py
```diff
@@ -185,6 +218,108 @@ def _quit() -> None:
...
+        local_filename = os.path.join(downloads_path, FILE_NAME)
+
+        response = requests.get(DOWNLOAD_URL, stream=True)
+        total_size = response.headers.get("content-length")
+        total_size = int(total_size) if total_size else None
+        block_size = 1024  # 1 Kilobyte
+        start_time = time.time()
+        last_update_time = start_time
+        with open(local_filename, "wb") as file, tqdm(
+            total=total_size, unit="B", unit_scale=True, desc=FILE_NAME
```
abrichr, line 268:
```suggestion
            total=total_size, unit="B", unit_scale=True, desc=file_name
```

abrichr:
Hi @shashank40 , thank you for your patience,

I tried running this on my machine:

```
open_adapt-v0.36.0.app.zip:  31%|██████████████████████████████▍                                                                   | 9.00/29.0 [00:00<00:00, 75.8kB/s]
Exception in thread Thread-5 (download_latest_version):
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.10/3.10.14/Frameworks/Python.framework/Versions/3.10/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/opt/homebrew/Cellar/python@3.10/3.10.14/Frameworks/Python.framework/Versions/3.10/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/abrichr/oa/OpenAdapt/openadapt/app/tray.py", line 297, in download_latest_version
    unzip_file(local_filename)
  File "/Users/abrichr/oa/OpenAdapt/openadapt/update_utils.py", line 19, in unzip_file
    shutil.unpack_archive(file_path, os.path.dirname(file_path))
  File "/opt/homebrew/Cellar/python@3.10/3.10.14/Frameworks/Python.framework/Versions/3.10/lib/python3.10/shutil.py", line 1314, in unpack_archive
    func(filename, extract_dir, **kwargs)
  File "/opt/homebrew/Cellar/python@3.10/3.10.14/Frameworks/Python.framework/Versions/3.10/lib/python3.10/shutil.py", line 1201, in _unpack_zipfile
    raise ReadError("%s is not a zip file" % filename)
shutil.ReadError: /Users/abrichr/Downloads/open_adapt-v0.36.0.app.zip is not a zip file
```

Any suggestions?

openadapt/app/tray.py
```diff
@@ -155,6 +169,25 @@ def __init__(self) -> None:
...
+        self.dashboard_action = TrackedQAction("Launch Dashboard")
+        self.dashboard_action.triggered.connect(self.launch_dashboard)
+        self.menu.addAction(self.dashboard_action)
+
+        self.download_button_text = (
+            f"Download Latest Version v{LATEST_VERSION}"
+            if Version(CURRENT_VERSION) < Version(LATEST_VERSION)
+            else "No updates available"
+        )
+        self.download_update_action = TrackedQAction(self.download_button_text)
```
abrichr, line 181: What do you think about disabling the menu item if no updates are available?

shashank40:
@abrichr Any more changes that might expect ?
I would complete it today so that we can get this pending PR in.

abrichr:
Hi @shashank40 , there should not be any more changes. I just noticed I introduced the latest bug with my changes 😅 will fix.

shashank40:
Yes @abrichr i saw that. It was actually working fine for me till that commit.
image


Even now works fine for me

shashank40:
@abrichr I have added started unzipping toast.
I have also added the check if i am on the latest version, menu action will be disabled.

shashank40:
@abrichr Please have a look at the latest code.

openadapt/app/tray.py
shashank40: Done

openadapt/app/tray.py
```diff
@@ -12,10 +12,13 @@
 import multiprocessing
 import os
 import sys
```
abrichr, line 14:
```suggestion
import requests
import sys
import threading
import time

```

openadapt/app/tray.py
```diff
@@ -12,10 +12,13 @@
...
-
+import requests
+import threading
+from tqdm import tqdm
+import time
```
abrichr, line 18:
```suggestion
from tqdm import tqdm
```

openadapt/app/tray.py
```diff
@@ -185,6 +226,123 @@ def _quit() -> None:
...
+                            f"Estimated time remaining: {eta_formatted} minutes"
+                        )
+                        self.download_start_toast.emit(progress_message)
+
+            self.unzipping_started_toast.emit("Unzipping Started")
+            unzip_file(local_filename)
+            self.download_complete.emit("Download & Unzipping Complete")
+        except Exception as e:
+            logger.error(e)
+            self.download_complete.emit("Download Failed")
```
abrichr, line 319: I would expect this message to be symmetrical to the success case, e.g.:

```
"Download & Unzipping Complete"
...
"Download & Unzipping Failed"
```

Perhaps instead of `Download & Unzipping` it can be simply `Update`

Also, should we rename `download_complete` to e.g. `update_complete`?

openadapt/update_utils.py
```diff
@@ -0,0 +1,33 @@
...
+
+def set_permissions(path: str) -> None:
+    """Set the permissions of all files to make the executable."""
+    for root, dirs, files in os.walk(path):
+        for dir in dirs:
+            dir_path = os.path.join(root, dir)
+            try:
+                os.chmod(os.path.join(root, dir), 0o755)
+            except PermissionError:
+                print(f"Skipping directory due to PermissionError: {dir_path}")
```
abrichr, line 15: Please use `logger` instead of `print`

openadapt/update_utils.py
```diff
@@ -0,0 +1,33 @@
...
+                print(f"Skipping file due to PermissionError: {file_path}")
+            except Exception as e:
+                print(f"An error occurred for file {file_path}: {e}")
+
+
+def unzip_file(file_path: str) -> None:
+    """Unzip a file to the given directory."""
+    if os.path.exists(file_path):
+        shutil.unpack_archive(file_path, os.path.dirname(file_path))
+        set_permissions(os.path.dirname(file_path))
```
abrichr, line 32: I do not believe it should be necessary to set permissions in the entire Downloads directory. Is this line even necessary? If so, can we create a new folder and set permissions on that instead?

openadapt/app/tray.py
```diff
@@ -107,9 +117,13 @@ class SystemTrayIcon:
 
     # storing actions is required to prevent garbage collection
     recording_actions = {"visualize": [], "replay": []}
+    download_complete = Signal(str)
+    download_start_toast = Signal(str)
+    unzipping_started_toast = Signal(str)
```
abrichr, line 122: What do you think about adding the suffix `_signal` to each of these?

abrichr (commented):
Thank you @shashank40 ! Almost there.

openadapt/app/tray.py
shashank40: Done

openadapt/update_utils.py
shashank40: Yess, that would be better. Thanks for the suggestion.

 I have made a few changes. Now the ZIP is extracted to a specified folder. This folder will have the name as OpenAdapt-v{latest_version}. The file will be unzipped inside this folder and the permission of this dir and file inside it will be changed. 

This makes it cleaner too as folder will have specified name with the version.

Something like this.
image

openadapt/update_utils.py
shashank40: Done

shashank40:
Hi @abrichr 
Made all the specified changes. Is this better?

openadapt/app/tray.py
shashank40: Did this with isort

openadapt/app/tray.py
shashank40: Did this with isort

shashank40:
@abrichr any update?

openadapt/app/tray.py
```diff
@@ -183,7 +225,125 @@ def _quit() -> None:
...
+        """Stops download when button clicked."""
+        self.cancel_download_event.set()
+        self.show_toast("Stopping download...", duration=4000)
+
+    @Slot(str)
+    def download_start_toast_slot(self, message: str) -> None:
+        """Shows download start toast depending on emitted signal."""
+        self.download_progress_toast = self.show_toast(
+            message,
+            duration=DOWNLOAD_TOAST_UPDATE_TIME * 1000 - 50,
```
abrichr, line 240: @shashank40 can you please clarify why `* 1000 - 50`? Ideally this would be clarified in code, e.g.

DOWNLOAD_TOAST_DURATION_MILLISECONDS = 3 * 1000 - 50  # subtract 50 for xyz reason

shashank40: Yess @abrichr , it was to handle condition between spitting new toast and old toast times out by then.
But i have changed it to .hide()

shashank40:
@abrichr Can you please do a final review?

abrichr:
Thank you @shashank40. There appear to be a few loose ends:

1. The "extracting" toast disappears, and there is no indication to the user that anything is happening. Ideally we show progress on this as well, e.g. using the `zip_file` module:

```
def extract_with_progress_zip(zip_path: str, extract_to: str):
    with zipfile.ZipFile(zip_path, 'r') as zip_ref:
        total_files = len(zip_ref.infolist())
        for i, file in enumerate(zip_ref.infolist()):
            zip_ref.extract(file, extract_to)
            logger.info(f"Extracted {i + 1} of {total_files} files ({(i + 1) / total_files * 100:.2f}%)")
            # TODO: update toast, like with downloading
```

2. Once unzipping is complete, the user will have more than one copy of OpenAdapt on their hard drive. Ideally we:
a. Move the newly extracted version to the same directory as the old version.
b. Ask the user whether they would like to remove the previous version, and if so then remove it*.

3. Finally, I would expect that:
a. the currently running of OpenAdapt is shut down
b. the newly extracted version's executable is executed

* The removing will probably need to be done on the first launch of the new version.

I would like to merge this now, but as it currently stands, it's not very user friendly without these additions, i.e. once the extraction is complete, the user likely expects to be running the new version, even though currently they won't be.

Once these additions have been implemented, the update feature will be complete and we can merge this.

Thank you @shashank40 , almost there!

shashank40:
@abrichr I think we should not do part 1), reason is extraction takes 5-10secs. Now we already show toast when it starts and when extraction completes. Now the overhead and resource consumption to just show extraction time(which in turn is shown by 2 different toast) seems excessive.

@abrichr i would request you to please review the PR all at once if you can. We have been going too and fro for too long and every time a new commit is pushed, there is a new change requested on the old code itself which could've been done earlier. It would make things easier for both of us.

Also to add, the new features requested is a bit out of scope of the initial aim of the project. I mean i am more than happy to do it but 3 weeks of too and fro and a new feature addition for just $100 bounty is not worth the effort.
If we can get the current PR in and then add these things as a new feature, we can do that.

abrichr:
Thank you @shashank40 , I appreciate you presenting an alternative perspective.

Regarding extracting the zip file: on my machine, it can take a while. The notification disappears before the extraction is complete, then it's not clear whether the process is complete or not. I think it's important we implement similar logic as for downloading to indicate the update is still occurring.

Regarding not removing the old application, the intention here is to emulate the behavior of other applications that update automatically. For example, when I update Chrome, I don't have to remove the old directory.

shashank40:
@abrichr When comparing it to chrome, browsers are installed applications which have their own directory on the hard-drive and change required files when updated. So they never touch any other directory/downloaded file. Even if we download a multiple zips for chrome, they are never deleted by chrome.

So installed apps work very differently in how updates work and how files are changed in their installed directory locations.

OpenAdapt is a desktop app and not and installed app. It doesn't have a fixed directory like installed apps and so deleting files from any other directory needs higher level permissions and makes it unsafe. If OpenAdapt can delete files from any directory, it makes it unsafe.

shashank40:
@abrichr updates?

abrichr:
Hi @shashank40 , thank you for your patience.

> OpenAdapt is a desktop app and not and installed app.

We've implemented installers in https://github.com/OpenAdaptAI/OpenAdapt/pull/858. We would like the update functionality to work with the installed app.

abrichr:

Issue #724, opened 2024-06-07
Upload data from dashboard


abrichr:
### Feature request

We would like to give users the option to share their recording data with us.

This involves:
- creating a server / serverless application to accept upload requests and store to s3
- creating a button in the dashboard to upload a scrubbed recording

### Motivation

https://www.oreilly.com/radar/what-we-learned-from-a-year-of-building-with-llms-part-iii-strategy/

![image](https://github.com/OpenAdaptAI/OpenAdapt/assets/774615/d94a5352-1f4c-487f-a015-7a1ab9e4936f)

abrichr:
https://aws.amazon.com/blogs/compute/uploading-to-amazon-s3-directly-from-a-web-or-mobile-application/

abrichr:
Closing as part of the architecture transition to modular packages. The legacy codebase has been frozen at v0.46.0.

Upload recordings from dashboard was for the legacy dashboard which has been replaced.

For the new modular architecture, please open issues in the relevant package repository:
- GUI capture: https://github.com/OpenAdaptAI/openadapt-capture
- ML/training: https://github.com/OpenAdaptAI/openadapt-ml
- Benchmarks: https://github.com/OpenAdaptAI/openadapt-evals
- Privacy: https://github.com/OpenAdaptAI/openadapt-privacy

See PR #960 for details on the new architecture.

PR #787, opened 2024-06-21
Upload recording feature

KIRA009:
This PR addresses #724 

**Summary**
This PR adds a script to deploy an app to AWS lambda, that is then used by the openadapt app to upload zipfiles of recordings from users.

**Checklist**

From the project root, run `python -m scripts.recording_uploader.deploy` (ensure that you have the necessary aws creds configured). Once the command completes, note the api url in the output, and paste that onto the `config.py`'s `RECORDING_UPLOAD_URL` variable. Once that is done, start the app, navigate to a recording detail page, and click on the "Upload recording" button. Check the s3 bucket to confirm that the recording has been uploaded (in the form of a zip file)

**Other information**

### DIFF
--- a/admin/recording_uploader/deploy.py
+++ b/admin/recording_uploader/deploy.py
@@ -51,7 +51,8 @@ def main(region_name: str = "us-east-1", destroy: bool = False) -> None:
             logger.info("Lambda function deleted successfully.")
         else:
             stdout = ret.stdout.decode("utf-8") if ret.stdout else ""
-            # find the url, which is in the format https://${ServerlessRestApi}.execute-api.${AWS::Region}.amazonaws.com/Prod/upload/
+            # find the url, which is in the format
+            # https://${ServerlessRestApi}.execute-api.${AWS::Region}.amazonaws.com/Prod/upload/
             url_match = re.search(
                 r"https://([^\.]+)\.execute-api\.([^\.]+)\.amazonaws\.com/Prod/upload/",
                 stdout,
--- a/openadapt/utils.py
+++ b/openadapt/utils.py
@@ -1132,7 +1132,10 @@ def get_recording_url(uploaded_key: str, uploaded_to_custom_bucket: bool) -> str
 def delete_uploaded_recording(
     recording_id: int, uploaded_key: str, uploaded_to_custom_bucket: bool
 ) -> str:
-    def inner():
+    """Delete a recording from cloud."""
+
+    def inner() -> None:
+        """Delete a recording from cloud."""
         if uploaded_to_custom_bucket:
             s3 = boto3.client(
                 "s3",
--- a/admin/recording_uploader/uploader/app.py
+++ b/admin/recording_uploader/uploader/app.py
@@ -1,7 +1,6 @@
 """Lambda function for generating a presigned URL for uploading a recording to S3."""
 
 from typing import Any
-from uuid import uuid4
 import json
 
 from botocore.client import Config
@@ -70,7 +69,7 @@ def get_presigned_url(data: dict) -> dict:
 
 
 def delete_object(data: dict) -> dict:
-    """Delete an object from the s3 bucket
+    """Delete an object from the s3 bucket.
 
     Args:
         data (dict): The data from the request.
--- a/openadapt/app/dashboard/api/recordings.py
+++ b/openadapt/app/dashboard/api/recordings.py
@@ -103,7 +103,7 @@ class RecordingsAPI:
 
     @staticmethod
     def delete_recording_on_cloud(recording_id: int) -> dict[str, bool]:
-        """Delete a recording from cloud"""
+        """Delete a recording from cloud."""
         session = crud.get_new_session(read_only=True)
         recording = crud.get_recording_by_id(session, recording_id)
         if recording.upload_status == Recording.UploadStatus.NOT_UPLOADED:
--- a/openadapt/models.py
+++ b/openadapt/models.py
@@ -101,6 +101,8 @@ class Recording(db.Base):
     )
 
     class UploadStatus(enum.Enum):
+        """Enum representing the upload status of a recording."""
+
         NOT_UPLOADED = "not_uploaded"
         UPLOADING = "uploading"
         UPLOADED = "uploaded"


scripts/recording_uploader/__init__.py
```diff
@@ -0,0 +1 @@
+"""Init file for the recording_uploader package."""
```
abrichr, line 1: If this is a package, should we move it outside of `scripts`?

scripts/recording_uploader/uploader/app.py
```diff
@@ -0,0 +1,40 @@
...
+    return {
+        "statusCode": 200,
+        "body": json.dumps(get_presigned_url()),
+    }
+
+
+def get_presigned_url() -> dict:
+    """Generate a presigned URL for uploading a recording to S3."""
+    bucket = "openadapt"
+    region_name = "us-east-1"
```
abrichr, line 22: What do you think about putting these in `NAMED_CONSTANTS` at the top of this file, and passing them in as default keyword arguments to `get_presigned_url`?

abrichr (changes requested):
Thank you for putting this together @KIRA009 ! Just left a few small comments, happy to chat about any of it if you like! 🙏 😄

scripts/recording_uploader/README.md
```diff
@@ -0,0 +1,87 @@
...
+* [PhpStorm](https://docs.aws.amazon.com/toolkit-for-jetbrains/latest/userguide/welcome.html)
+* [PyCharm](https://docs.aws.amazon.com/toolkit-for-jetbrains/latest/userguide/welcome.html)
+* [RubyMine](https://docs.aws.amazon.com/toolkit-for-jetbrains/latest/userguide/welcome.html)
+* [DataGrip](https://docs.aws.amazon.com/toolkit-for-jetbrains/latest/userguide/welcome.html)
+* [VS Code](https://docs.aws.amazon.com/toolkit-for-vscode/latest/userguide/welcome.html)
+* [Visual Studio](https://docs.aws.amazon.com/toolkit-for-visual-studio/latest/user-guide/welcome.html)
+
+## Deploy the application
+
+There is a `deploy` script that creates the s3 bucket and deploys the application using the SAM CLI (included as part of the dev dependencies of this project). The bucket name is hardcoded in the script. The SAM CLI is set up to run in `guided` mode, which will prompt the user every time befor deploying, in case the user wants to change the default values.
```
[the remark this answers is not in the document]
abrichr: Can we make the bucket name configurable, with the aws region etc?

scripts/recording_uploader/deploy.py
```diff
@@ -0,0 +1,47 @@
...
+        proceed = input(f"Bucket '{bucket}' already exists. Proceed? [y/N] ")
+        if proceed.lower() != "y":
+            return
+
+    # deploy the code to AWS Lambda
+    commands = ["sam", "deploy"]
+    if guided:
+        commands.append("--guided")
+    subprocess.run(commands, cwd=CURRENT_DIR, check=True)
+    print("Lambda function deployed successfully.")
```
abrichr, line 43: What do you think about using `logger` here?

scripts/recording_uploader/deploy.py
```diff
@@ -0,0 +1,47 @@
...
+    bucket = "openadapt"
+
+    # create the S3 bucket, if it doesn't already exist
+    try:
+        s3.create_bucket(
+            ACL="private",
+            Bucket=bucket,
+        )
+    except (s3.exceptions.BucketAlreadyExists, s3.exceptions.BucketAlreadyOwnedByYou):
+        proceed = input(f"Bucket '{bucket}' already exists. Proceed? [y/N] ")
```
abrichr, line 34: Is this necessary? What happens if the user proceeds if the bucket has already been created?

scripts/recording_uploader/uploader/app.py
```diff
@@ -0,0 +1,40 @@
...
+    )
+    key = f"recordings/{uuid4()}.zip"
+
+    presigned_url = s3.generate_presigned_url(
+        ClientMethod="put_object",
+        Params={
+            "Bucket": bucket,
+            "Key": key,
+        },
+        ExpiresIn=3600,
```
abrichr, line 37: Can you please make this a named constant, e.g. `ONE_HOUR_IN_SECONDS`?

scripts/recording_uploader/uploader/app.py
```diff
@@ -0,0 +1,40 @@
...
+    """Generate a presigned URL for uploading a recording to S3."""
+    bucket = "openadapt"
+    region_name = "us-east-1"
+    s3 = boto3.client(
+        "s3",
+        config=Config(signature_version="s3v4"),
+        region_name=region_name,
+        endpoint_url=f"https://s3.{region_name}.amazonaws.com",
+    )
+    key = f"recordings/{uuid4()}.zip"
```
abrichr, line 29: What do you think about adding the user's unique id to the path, e.g. `recordings/{user_id}/{upload_id}.zip`?

abrichr:
Once we start scaling we should consider supporting B2, e.g. app.py:

```
"""Lambda-like function for generating a presigned URL for uploading a recording to B2."""

from typing import Any
from uuid import uuid4
import json
from b2sdk.v2 import B2Api, InMemoryAccountInfo

def get_b2_client() -> B2Api:
    """Create and return a B2 client."""
    info = InMemoryAccountInfo()
    b2_api = B2Api(info)
    b2_api.authorize_account("production", "applicationKeyId", "applicationKey")
    return b2_api

def lambda_handler(*args: Any, **kwargs: Any) -> dict:
    """Main entry point for the function."""
    return {
        "statusCode": 200,
        "body": json.dumps(get_presigned_url()),
    }

def get_presigned_url() -> dict:
    """Generate a presigned URL for uploading a recording to B2."""
    bucket_name = "openadapt"
    b2_api = get_b2_client()
    bucket = b2_api.get_bucket_by_name(bucket_name)
    file_name = f"recordings/{uuid4()}.zip"
    file_info = {'how': 'good-file'}

    presigned_url = bucket.get_upload_url(file_name, file_info=file_info)
    
    return {"url": presigned_url['upload_url'], "upload_auth_token": presigned_url['authorization_token']}

```

For now let's stick with s3.

scripts/recording_uploader/__init__.py
KIRA009: It should be named as a module instead 😅

scripts/recording_uploader/deploy.py
KIRA009: I have removed this

KIRA009:
Before this is merged, we need to setup the upload url and update in on `config.py` - `RECORDING_UPLOAD_URL`

scripts/recording_uploader/deploy.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def main(region_name: str = "us-east-1", guided: bool = True) -> None:
+    """Deploy the uploader to AWS Lambda.
+
+    Args:
+        region_name (str): The AWS region to deploy the Lambda function to.
+        guided (bool): Whether to use the guided SAM deployment.
+    """
+    s3 = boto3.client(
```
abrichr, line 21: @KIRA009 can you please modify this to use the credentials specified in `openadapt.config`? Specifically we should add `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` or similar.

Also, please add some documentation regarding the permissions required for this IAM user.

scripts/recording_uploader/uploader/requirements.txt
```diff
@@ -0,0 +1 @@
+boto3==1.34.84
```
abrichr, line 1: What do you think about moving all of this outside of `/scripts`, and into e.g. `/admin` or similar?

scripts/recording_uploader/deploy.py
KIRA009: Do we want to add these keys to the config? They won't be used anywhere else in the project, and I think when you run the deploy script, if boto3 doesn't find appropriate keys in the place where its looking, the user is notified of that.

For the access keys, I followed this without any explicit permissions - https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html#Using_CreateAccessKey. These keys are the ones that you need to run the deploy script, which then creates appropriate iam users on its own. Once its deployed, if you want you can delete the original access keys. Should I add this in the readme?

scripts/recording_uploader/uploader/requirements.txt
KIRA009: I think the `uploader` module has to remain inside the `recording_uploader` module. If you are talking about the `recording_uploader` module, we could, but I think `scripts` is also a good enough place for it to be in. Let me know

abrichr: Yes I meant to move `recording_uploader` into a new directory, `/admin` or similar.

openadapt/share.py
```diff
@@ -173,6 +174,37 @@ def receive_recording(wormhole_code: str) -> None:
...
+            )
+
+            # Delete the zip file after uploading
+            if os.path.exists(zip_file_path):
+                os.remove(zip_file_path)
+                logger.info(f"deleted {zip_file_path=}")
+        except Exception as exc:
+            logger.exception(exc)
+
+    Thread(target=_inner).start()
```
abrichr, line 205: Should we add some error handling / retrying / reporting around this?

openadapt/utils.py
```diff
@@ -946,6 +948,23 @@ def get_posthog_instance() -> DistinctIDPosthog:
...
 
 
+def upload_file_to_s3(file_path: str, body: dict) -> requests.Response:
+    """Upload a file to an S3 bucket.
+
+    Args:
+        file_path (str): The path to the file to upload.
+        body (dict): The body of the request.
+    """
+    filename = os.path.basename(file_path)
```
abrichr, line 958:
```suggestion
    file_name = os.path.basename(file_path)
```

abrichr: What do you think about logging this path?

abrichr (changes requested):
Thank you for putting this together @KIRA009 !

What do you think about something like this:
```
class Recording(db.Base):
    __tablename__ = "recording"
...
    s3_url = sa.Column(sa.String, nullable=True)
```

abrichr:
What do you think about renaming the directory to simply `uploader` instead of `recording_uploader`?

abrichr:
@KIRA009 what do you think about https://www.cloudflare.com/en-ca/developer-platform/products/r2/:

> Forever Free
> 10 GB / month

admin/recording_uploader/deploy.py
```diff
@@ -10,31 +12,64 @@
...
-    # deploy the code to AWS Lambda
-    commands = ["sam", "deploy"]
-    if guided:
-        commands.append("--guided")
-    subprocess.run(commands, cwd=CURRENT_DIR, check=True)
-    logger.info("Lambda function deployed successfully.")
+    # check if aws credentials are set
+    if os.getenv("AWS_ACCESS_KEY_ID") is None:
+        raise ValueError("AWS_ACCESS_KEY_ID is not set")
+    if os.getenv("AWS_SECRET_ACCESS_KEY") is None:
```
abrichr, line 25: Why not read this from `config`?

admin/recording_uploader/deploy.py
```diff
@@ -10,31 +12,64 @@
...
+        raise ValueError("AWS_SECRET_ACCESS_KEY is not set")
+    if destroy:
+        commands = ["sam", "delete", "--no-prompts"]
+    else:
+        s3 = boto3.client(
+            "s3",
+            region_name=region_name,
+            endpoint_url=f"https://s3.{region_name}.amazonaws.com",
+        )
+        bucket = "openadapt"
```
abrichr, line 35: What do you think about defining this is config.py?

admin/recording_uploader/deploy.py
KIRA009: This script is not supposed to be part of the OpenAdapt app, its an admin script that needs to be run by the owner (you) on a machine that has the relevant aws creds in its environment. From the PR description

> From the project root, run python -m scripts.recording_uploader.deploy (ensure that you have the necessary aws creds configured). Once the command completes, note the api url in the output, and paste that onto the config.py's RECORDING_UPLOAD_URL variable

Because config.py is more closely related to settings of the app, I didn't think it'd be useful to add these there.

admin/recording_uploader/deploy.py
KIRA009: The same reason as above, plus this is hardcoded because this script will be run once in a while (only when the lambda function is changed). And ideally we won't be changing bucket names between runs.

openadapt/utils.py
KIRA009: Done

openadapt/share.py
KIRA009: Added retries

admin/recording_uploader/deploy.py
abrichr: I think we want to read from config.py.

KIRA009: How would you suggest a user override these settings? The default values will be empty in `config.py` and `config.defaults.json`, so the only ways are to manually edit the `config.json` file in the `data` folder, or we expose it in the dashboard settings page? A regular user won't be needing to edit this, and might get confused if its in the dashboard settings

abrichr:

Issue #804, opened 2024-06-26
Store application logs


abrichr:
### Feature request

We would like to store application logs to a file.

### Motivation

Debugging

KIRA009:
This has been addressed

PR #810, opened 2024-06-27
Override std out for application

KIRA009:
This PR addresses #804 

**Summary**
This PR wraps the entire app in a context processor, that redirects all stdout and stderr to a log file

**Checklist**

Build the app, then try and run it. You should be seeing 

**Other information**

### DIFF
--- a/openadapt/build_utils.py
+++ b/openadapt/build_utils.py
@@ -1,8 +1,9 @@
 """Utility functions for the build process."""
 
-import os
+import importlib.metadata
 import pathlib
 import sys
+import time
 
 
 def get_root_dir_path() -> pathlib.Path:
@@ -36,7 +37,7 @@ class RedirectOutput:
     def __enter__(self) -> "RedirectOutput":
         """Redirect stdout and stderr to /dev/null."""
         if is_running_from_executable():
-            log_file_path = get_root_dir_path() / "data" / "openadapt.log"
+            log_file_path = get_log_file_path()
             log_stream = open(log_file_path, "a")
             self.old_stdout = sys.stdout
             self.old_stderr = sys.stderr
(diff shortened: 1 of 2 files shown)


openadapt/build_utils.py
```diff
@@ -36,10 +36,11 @@ class RedirectOutput:
     def __enter__(self) -> "RedirectOutput":
         """Redirect stdout and stderr to /dev/null."""
         if is_running_from_executable():
-            null_stream = open(os.devnull, "a")
+            log_file_path = get_root_dir_path() / "data" / "openadapt.log"
```
abrichr, line 39: What do you think about adding the date/time to the filename?

openadapt/entrypoint.py
```diff
@@ -13,21 +13,19 @@
...
+            from openadapt.config import print_config
 
-        print_config()
-        load_alembic_context()
-        tray._run()
-    except Exception as exc:
-        data_dir = get_root_dir_path()
-        # TODO: log all exceptions to a file
-        with open(data_dir / "error.log", "a") as f:
-            f.write(f"{datetime.now()}: {exc}\n")
```
abrichr, line 28: Why was this removed? Why not log to the same file? e.g.

```
with redirect_stdout_stderr() as redirect:
    try:
        ...
    except:
        with open(redirect.log_file, "a") as f:
            ...
```

abrichr (commented):
Thank you @KIRA009 ! Just a couple of small comments

openadapt/entrypoint.py
KIRA009: Since we are overriding stdout, print will automatically write to that file

openadapt/build_utils.py
KIRA009: Makes sense. I was also wondering if we shpuld separate log files by versions?

abrichr: I think as long as the version is logged, date/time in the path is sufficient.

openadapt/entrypoint.py
abrichr: Why `print` and not `logger.info`?

openadapt/entrypoint.py
```diff
@@ -13,21 +13,19 @@
...
+            print_config()
+            load_alembic_context()
+            tray._run()
+        except Exception as exc:
+            # TODO: log all exceptions to a file
+            print(f"{datetime.now()}: {exc}")
```
abrichr, line 27: Why `print` and not `logger.exception(exc)`?

openadapt/build_utils.py
KIRA009: Should be easy, and also should help when in cases where the log file gets too large

abrichr (approved):
LGTM!

PR #812, opened 2024-06-27
Fix tray signals

KIRA009:
This PR fixes the issues with signals not being passed around properly on windows

**Summary**
This PR replaces the QSocketNotifier with a QThread and a worker that uses a signal in addition to a parent/child pipe because in windows to pass data around. This additional maneouver was needed because windows wasn't triggered the activated signal of the qsocketnotifier object. 

The current data flow is: `child_conn` -> `parent_conn` -> `signal emit` -> `signal emit listener`.

The signal emit listener is needed, because processing things in the Qthread object is not possible because of how Qt restricts updates from amywhere else other than the main thread. The signal emit listener is executed on the main thread, so the signal is used as a connection to send the data back to the main thread.

**Checklist**

Run the code (in development mode) in windows, and ensure that you see the notifications and the tray text changes.

**Other information**

### DIFF
--- a/.github/workflows/release-and-publish.yml
+++ b/.github/workflows/release-and-publish.yml
@@ -140,7 +140,7 @@ jobs:
         id: test_on_windows
         shell: powershell
         run: |
-          ./build_scripts/test_app_run_windows.bat
+          cmd.exe /c .\build_scripts\test_app_run_windows.bat
           if ($LastExitCode -ne 0) {
             "windows_build_status=failed" | Out-File -FilePath $env:GITHUB_OUTPUT -Append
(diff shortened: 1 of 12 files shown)


abrichr:
Thank you @KIRA009 ! Please address the build issues when you get a chance so we can get this merged.

Also I noticed this includes the changes from https://github.com/OpenAdaptAI/OpenAdapt/pull/810/files, which itself has outstanding comments. How do you suggest we proceed?

abrichr (approved):
Thank you @KIRA009 🙏 
LGTM!

PR #820, opened 2024-07-02
cursorReplay

Ahmadkhan02:
**Checklist**

### DIFF
--- a/openadapt/prompts/generate_action_event_with_red_dot.j2
+++ b/openadapt/prompts/generate_action_event_with_red_dot.j2
@@ -1,0 +1,33 @@
+Consider the previously recorded actions:
+
+```json
+{{ recorded_actions }}
+```
+
+Consider the actions you've produced (and we have replayed) so far:
+
+```json
+{{ replayed_actions }}
+```
+
+Consider the user's proposed modifications in natural language replay instructions:
+
+```text
+{{ replay_instructions }}
+```
+
+Consider the current active window:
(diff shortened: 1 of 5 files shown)


Ahmadkhan02:
Hi @abrichr , I was a bit confused and had some questions regarding the issue.

1. Should self correction and adding a red dot be seperate action_events? or can do do drawing of red dot and return self correction as the action event (current implementation)
2.  I have no idea as to what action_event should be returned. Should the strategy return either red dot drawing event or self correction event depending on the state?
3. Does capability for playback etc for drawing red dot ?

I am not quite able to frame questions well enough to clear my doubts. It would be quite helpful if you could give me an idea as to how this strategy would be used and whats expected from it. I get the stuff about mimicing the workflow of humans solving mathematics as an example but wanted to know how we need that here.

abrichr:
Hi @Ahmadkhan02 , thank you for your interest!

> Should self correction and adding a red dot be seperate action_events? or can do do drawing of red dot and return self correction as the action event (current implementation)

Adding a red dot should not involve any action_events.

> I have no idea as to what action_event should be returned. Should the strategy return either red dot drawing event or self correction event depending on the state?

The only action_event(s) that should be returned are the ones that should be played back on the user's machine in order to accomplish their instructions.

I recommend starting with an easier task. This one appears to be close to solved in https://github.com/OpenAdaptAI/OpenAdapt/pull/823.

Issue #841, opened 2024-07-08
[Bug]: Nonetype object has no attribute write


abrichr:
### Describe the bug

```
Traceback (most recent call last):
  File "sqlalchemy\engine\base.py", line 1900, in _execute_context
  File "sqlalchemy\engine\default.py", line 736, in do_execute
sqlite3.InterfaceError: Error binding parameter 5 - probably unsupported type.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "multiprocessing\process.py", line 314, in _bootstrap
  File "multiprocessing\process.py", line 108, in run
  File "openadapt\utils.py", line 727, in wrapper_logging
    result = func(*args, **kwargs)
  File "openadapt\record.py", line 387, in write_events
    state = write_fn(session, recording, event, perf_q, **(state or {}))
  File "openadapt\record.py", line 315, in write_window_event
    crud.insert_window_event(db, recording, event.timestamp, event.data)
  File "openadapt\db\crud.py", line 154, in insert_window_event
    _insert(session, event_data, WindowEvent, window_events)
  File "openadapt\db\crud.py", line 80, in _insert
    result = session.execute(sa.insert(table), to_insert)
  File "sqlalchemy\orm\session.py", line 1714, in execute
  File "sqlalchemy\engine\base.py", line 1705, in _execute_20
  File "sqlalchemy\sql\elements.py", line 333, in _execute_on_connection
  File "sqlalchemy\engine\base.py", line 1572, in _execute_clauseelement
  File "sqlalchemy\engine\base.py", line 1943, in _execute_context
  File "sqlalchemy\engine\base.py", line 2124, in _handle_dbapi_exception
  File "sqlalchemy\util\compat.py", line 208, in raise_
  File "sqlalchemy\engine\base.py", line 1900, in _execute_context
  File "sqlalchemy\engine\default.py", line 736, in do_execute
sqlalchemy.exc.InterfaceError: (sqlite3.InterfaceError) Error binding parameter 5 - probably unsupported type.
[SQL: INSERT INTO window_event (id, recording_timestamp, recording_id, timestamp, state, title, "left", top, width, height, window_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)]
[...]

PR #854, opened 2024-07-16
Application logs

KIRA009:
This PR addresses the root cause of #841 

**Summary**

This PR runs all `multiprocessing.Process`targets inside a try catch block, so any errors caused from within the target are gracefully handled.

The second part of this PR modifies the logger used by all processes to replace the default handler with a file handler, so all logs (including any exception traces) are written to the log file and can be used for debugging.

**Checklist**

Introduce an exception in one of the crud methods (e.g. `insert_window_event`) and then build the app (`python -m openadapt.build`) and run the built app. Start and complete a recording. The app should not crash and the logs should contain the exception raised in the crud method. The logs will be available at the `~/Library/Preferences/openadapt/data` in Macos and `AppData/Roaming/openadapt` in Windows

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ from pprint import pformat
 import json
 
 # external
-from loguru import logger
+from openadapt.custom_logger import logger
 import numpy as np
 
 # local
--- a/experiments/fastsamsom.py
+++ b/experiments/fastsamsom.py
@@ -2,12 +2,11 @@
 
 from pprint import pformat
 
-from loguru import logger
 from PIL import Image
 import numpy as np
 
 from openadapt import adapters, config, contrib, utils, vision
-
+from openadapt.custom_logger import logger
 
 CONTRAST_FACTOR = 10000
 DEBUG = False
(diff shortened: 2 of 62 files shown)


KIRA009:
The second commit is entirely about replacing uses of the loguru logger with the custom logger, so the only commit that has functional changes is the first one - 4307a65bb0d95c911410dd7594efc1efc08b3f41

abrichr:
Thank you @KIRA009 ! Have you tested this on the built application and confirmed it's working as intended? Can you please show some screenshots confirming this to be the case?

KIRA009:
@abrichr Yes, I have tested it on mac. My windows machine is under repair so I am not able to test on that for the moment.

https://github.com/user-attachments/assets/bf9a2629-bbd2-4287-abbf-9dc6a0a7438a

abrichr:
Thank you. Why does it say "Window events are not yet supported"?

abrichr:
Thank you. What do you think about adding a test to assert that the application log is created and is non-empty?

Also, why does it say "Window events are not yet supported"?

KIRA009:
@abrichr That is a custom exception that I raised just to simulate an exception. The exception message is not real and not related to the app

PR #858, opened 2024-07-17
modify build.py to produce installers

abrichr:
We would like to implement automatic updates. Without installing to a system directory, however, this is not user friendly (see https://github.com/OpenAdaptAI/OpenAdapt/pull/782#issuecomment-2233621865).

This PR implements functionality to produce an installer so that the installed application can be updated more easily.

### DIFF
--- a/.github/workflows/release-and-publish.yml
+++ b/.github/workflows/release-and-publish.yml
@@ -47,14 +47,11 @@ jobs:
         - name: Build executables
           run: |
               poetry run python -m openadapt.build
-              cd dist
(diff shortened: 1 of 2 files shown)


KIRA009:
This looks good, tested on Mac. We might also have to change the ci scripts, unless the dmg making process is using stuff that comes pre-installed with the mac image

abrichr:
@KIRA009 I believe we should modify the Github actions to remove the zipping functionality, since Inno Setup already seems to compress the setup file, and I believe .dmg files are also compressed. What do you think?

PR #940, opened 2025-01-26
[draft] feat: dockerize

bard:

Ability to create a browser-accessible Docker image where OpenAdapt is set up and ready to run.

https://github.com/user-attachments/assets/64df5dcc-063a-476c-941c-e07194e2d65c

**Summary**

Goals:

1. cut the effort to try out OpenAdapt down to a single command
2. run OpenAdapt on Linux workstations where the snowball effect of replacing key components of the OS would make it impractical (e.g. Wayland→X11, pipewire→portaudio, Qt6.x→Qt6.8)
3. run OpenAdapt worry-free because the environment is isolated
4. run OpenAdapt on common cloud infrastructure, whether for production or CI

**Checklist**

Launch with:

```sh
docker build -t openadapt-local .
docker run --rm --shm-size=512m -p 6901:6901 openadapt-local
```

Then open http://localhost:6901 in the browser.

**Other information**

How it works:

- uses Ubuntu jammy (from the [kasm project](https://github.com/kasmtech/)), as it comes by default with a supported python
- installes build and runtime dependencies (node21, tesseract, portaudio)
- installs cpu-only torch and torchvision
- installs cpu-only [prebuilt detectron](https://github.com/MiroPsota/torch_packages_builder)
- moves `ForceFloat` out of `models.py` because requiring models.py in migrations otherwise triggers an import chain that ends with packages that expect an X11 environment, which is not available during the docker build

There are probably more correct ways for doing all of the above; for now I've only aimed at the first step of the "make it work, make it right, make it fast" mantra.

What works:

- `python -m openadapt.record "testing out openadapt"`
- `python -m openadapt.app.dashboard.run`

What doesn't yet:

- commands that rely on PySide, such as `python -m openadapt.entrypoint` (downgrading PySide to 6.2 to match system-installed Qt libraries makes these commands run, but also causes segfaults on `openadapt.record`)


abrichr:

PR #950, opened 2025-05-13
chore: add recorder extra to pyproject.toml

Omokami:


pyproject.toml
```diff
@@ -162,3 +162,22 @@ upload_to_release = true
...
+    "fire",
+    "loguru",
+    "soundfile",
+    "multiprocessing-utils",
+    "screen-recorder-sdk",      # Windows only
+    "pyaudio",                  # Windows/Linux
+    "pywin32",                  # Windows only
+    "oa-atomacos",              # macOS only
+    "pyobjc-framework-avfoundation", # macOS only
+    "xcffib"                    # Linux only
```
abrichr, line 182: Can we use markers here, like in e.g. https://github.com/OpenAdaptAI/OpenAdapt/blob/1d69373c4db9ddcdefed8ff145cb652c2d5c2931/pyproject.toml#L26 ?

abrichr (changes requested):
Thank you for getting started on this @Omokami !

Can you please update the description with something meaningful?
e.g. what is this, how to use it, example log output

pyproject.toml
Omokami: No, you cannot use environment markers (like `markers = "sys_platform == 'win32'"`) inside the `[tool.poetry.extras]` section of `pyproject.toml`. 
The `extras` section is just a list of dependency names (as strings) that must match the names in your main `[tool.poetry.dependencies]` section.
 Markers are only supported in the main dependencies section.

Omokami:
## What is this?

This PR adds a new Poetry extra called `recorder` to the `pyproject.toml` file. This extra includes only the minimal dependencies required to use the high-performance recorder functionality in `openadapt.record` (screen capture, keyboard/mouse logging, etc.), without installing the full set of OpenAdapt dependencies.

## How to use

After this change is merged, users can install OpenAdapt with just the recorder dependencies by running:

```bash
pip install openadapt[recorder]
```

This will install only the packages needed for the recorder, making it easier and faster to set up environments focused on recording functionality.

## Example usage

To start the recorder after installing with the `recorder` extra:

```bash
python -m openadapt.record
```

PR #1004, opened 2026-06-25
Fix GlitchTip error-reporting noise + import-smoke CI

abrichr:
## Why

GlitchTip (project `openadaptai/openadapt`, 3798) was dominated by **unactionable import-time crashes from automated/headless environments** (bot servers `admin123-nexus-*`, CI, no-display Linux) running OpenAdapt where it structurally can't run. These flood the dashboard and bury real issues. Triaging all 17 unresolved issues (65 events): ~80% of events are environmental noise, several "top bugs" are already fixed on this branch (pywinauto guarded, no `declarative_base(bind=)`) and recur only from old-release bots.

## What

- **`error_reporting.before_send_event`** now **drops unsupported-environment errors** (no display / `$DISPLAY` / X connection, wrong platform, `No module named 'pywinauto'`, numpy ABI `_ARRAY_API`, headless `crop_active_window`). It also **only shows the user alert when a GUI is actually running** — previously it popped a `QMessageBox` on *every* event inside the Sentry hook (broken/blocking headless).
- **`config.ERROR_REPORTING_DSN`** → live project **3798**. Previously **8771**, which isn't in the org and is unreachable, so its events were never visible in the dashboard.
- **CI import-smoke step** (`main.yml`): core modules must import cleanly — the dominant GlitchTip error class (same class as #999). TODO left for a Windows matrix (pywinauto path).
- **`gitignore *.pem`** — there was an untracked private key (`deploy/omniparser-key.pem`) at risk of `git add -A`.

## Validation

Ran the new filter against the **17 live issues**: it drops **53/65 events** (all the high-volume environmental flood) and keeps the **12-event actionable long tail**. At runtime it catches even more (the live hook sees full exception messages).

## Not done (deliberate)

- No `try/except` around `posthog`/`dictalchemy3`/`click` — they're *declared deps* (partial-install artifacts ≠ code bugs; `posthog` is subclassed at module scope). The import-smoke CI is the correct guard.
- No `numpy<2` pin (harmful in 2026; the one stale ABI error is filtered).

## Dashboard

Separately muted (status=ignored) the 7 high-volume noise issues (53 events) via the API — `ignored` not `resolved`, since old-release bots would reopen `resolved`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1005, opened 2026-06-25
Add crash reporting to the CLI (frozen-gated, filtered) + fix latent config dep

abrichr:
## Context

Error reporting was **removed in `c1efbece` — \"refactor: Convert to meta-package architecture (#960)\"**, which deleted `openadapt/error_reporting.py` and dropped the `sentry` dependency. It looks like collateral damage of the meta-package split, not a deliberate decision — so this re-adds it, built for the **new** thin-meta-package architecture.

(The GlitchTip dashboard noise that prompted this is from *old external deployments* running pre-refactor versions; separately triaged/muted. This PR is about restoring crash visibility for the current app, done right.)

## What

- **`openadapt/error_reporting.py`** — initializes Sentry/GlitchTip from the CLI entry point.
  - **Gated on `sys.frozen` (packaged build) by default** — *not* the old fragile `active_branch == "main"` check. So dev / source / CI / headless runs never report (that source-run flooding is exactly what buried real issues, e.g. the `admin123-nexus` headless containers).
  - **Master switch** `OPENADAPT_ERROR_REPORTING_ENABLED=0`; **force from source** `OPENADAPT_ERROR_REPORTING_FORCE=1`.
  - **`before_send` filter** drops unactionable unsupported-environment errors (no display, wrong platform, broken native deps).
  - **Never raises**; no-op if `sentry-sdk` isn't installed.
- **`config.py`** — `OPENADAPT_ERROR_REPORTING_{ENABLED,FORCE,DSN,ENVIRONMENT}` settings; DSN defaults to GlitchTip `openadaptai/openadapt` (3798). (DSN public key is an ingestion key, not a secret.)
- **`cli.py`** — `configure_error_reporting()` in the `main()` group, covering all subcommands.
- **`pyproject.toml`** — add `sentry-sdk`; **also add `pydantic-settings`**, a *latent missing base dep* — `config.py` imported it but it wasn't declared, so a clean `pip install openadapt` couldn't `import openadapt.config` (masked in CI by a transitive install).
- **`tests/test_error_reporting.py`** — filtering + gating (8 tests, no network).

## Validated

Clean-base venv install → `import openadapt.config` / `openadapt.error_reporting` OK, CLI import OK, 8 new tests pass, existing `test_cli_smoke` / `test_import_integrity` unaffected.

## How to test reporting end-to-end
`OPENADAPT_ERROR_REPORTING_FORCE=1 openadapt <cmd-that-errors>` → event appears in GlitchTip 3798 (filtered for env noise).

## Note
Supersedes #1004 (which targeted the abandoned `feat/deploy` branch — main had a different/removed structure). Closing that one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/openadapt/cli.py
+++ b/openadapt/cli.py
@@ -38,7 +38,11 @@ def main():
         openadapt train --capture my-task        # Train a model
         openadapt eval --checkpoint model.pt     # Evaluate the model
     """
-    pass
+    # Initialize crash reporting before dispatching to any subcommand. Lazy
+    # import keeps CLI startup fast and import-safe; the call never raises.
+    from openadapt.error_reporting import configure_error_reporting
+
+    configure_error_reporting()
 
 
 # =============================================================================
--- a/openadapt/config.py
+++ b/openadapt/config.py
@@ -50,6 +50,20 @@ class OpenAdaptSettings(BaseSettings):
     lambda_api_key: Optional[str] = None
 
     # =========================================================================
+    # Crash reporting (see openadapt/error_reporting.py)
+    # =========================================================================
+    # Master switch (OPENADAPT_ERROR_REPORTING_ENABLED).
+    error_reporting_enabled: bool = True
+    # Report from a source run too (default: only packaged/frozen builds report).
+    error_reporting_force: bool = False
+    # Sentry-compatible DSN; defaults to the GlitchTip openadaptai/openadapt (3798)
+    # project. The DSN public key is an ingestion key, not a secret.
+    error_reporting_dsn: Optional[str] = (
+        "https://dcf5d7889a3b4b47ae12a3af9ffcbeb7@app.glitchtip.com/3798"
+    )
+    error_reporting_environment: str = "production"
+
+    # =========================================================================
     # Paths
     # =========================================================================
     # Default directory for capture recordings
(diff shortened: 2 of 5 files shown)


abrichr:
**Triage 2026-07-17 — assessed against the post-#1012/#1013/#1014 launcher. Recommend CLOSE (superseded by the architecture, not by other code).**

Rationale:
1. **The gate can never fire in the current distribution.** Reporting is (correctly) gated on `sys.frozen`, but OpenAdapt now ships as wheel + sdist only (1.6.x releases have no frozen binaries), so the shipped artifact would never report — the PR is a no-op for users except via the FORCE env var.
2. **It cuts against the thin canonical-launcher role.** Post-#1013/#1014, runtime deps are deliberately just `click` + `openadapt-flow`; adding `sentry-sdk` (+ `pydantic-settings`) as hard deps of the meta-launcher expands the supply-chain surface for a wire that can't fire.
3. **The branch is 3 weeks stale against a rewritten `cli.py`/`config.py`/`pyproject`** (CONFLICTING); a rebase would be a substantial rewrite.
4. **Where crash reporting will actually pay off is the frozen desktop bundle** — the openadapt-desktop Tauri shell + PyInstaller engine sidecar. The frozen-gated + filtered design from this PR is the right template to port there when those builds go out to users.

Salvageable nugget worth a tiny separate PR if desired: `openadapt/config.py` on main still imports `pydantic_settings`, which is not a declared dependency — either declare it or make the import lazy/optional (nothing in the CLI imports `openadapt.config` today, so it only breaks direct imports on a clean install).

PR #1122, opened 2026-08-26
fix: make openadapt quickstart the successful first-run path

abrichr:
## What changed

`openadapt quickstart` remains the only command shown in the first-run docs and success output. The launcher now sends users from that result to the first-workflow guide.

The advanced `--simulate-rejected-write` option names the exact condition it tests. The launcher forwards that option to Flow. `--break-it` remains a hidden deprecated alias and prints a warning, so existing scripts don't fail without notice.

This draft depends on Flow accepting `--simulate-rejected-write` and writing its report to `run-rejected-write/REPORT.md`. It must not merge before that behavior is available in the supported Flow release.

## Checks

- Full tests: 197 passed, 6 skipped
- Ruff check and format check passed
- The wheel and source archive built successfully
- The public source-boundary check passed

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -10,7 +10,7 @@ installed `openadapt-flow` engine.
 ## First run
 
 ```bash
-openadapt quickstart [--headed] [--break-it] [--out NEW_DIRECTORY]
+openadapt quickstart [--headed] [--out NEW_DIRECTORY]
 ```
 
 `quickstart` runs the bundled synthetic workflow from recording through an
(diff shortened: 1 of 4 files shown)


abrichr:
Closing this one. The work is good and the rename is the right call, but it cannot merge yet, and this PR says so itself: *"It must not merge before that behavior is available in the supported Flow release."*

That condition is not met, and nothing is in flight to meet it.

**Two hard dependencies on Flow, both unmet:**

1. The flag. This forwards `--simulate-rejected-write` to the engine. `openadapt-flow` has zero occurrences of that string — its CLI is `--break-it` throughout (`openadapt_flow/__main__.py`), and it even prints `openadapt-flow tutorial --break-it` to users. Latest release is 1.33.0; this repo pins `openadapt-flow>=1.29.0,<2.0.0`.

2. The evidence path. This prints `run-rejected-write/REPORT.md`, but `openadapt_flow/tutorial.py:657` writes `run_dir=root / "run-broken"`.

So merging today breaks the advanced quickstart path twice over — and it breaks it for `--break-it` users too, since the deprecated alias now maps onto the new flag.

**Why the green suite did not catch this.** All three tests monkeypatch `_invoke_flow`, so they assert the launcher's argv and never execute Flow. "197 passed" is real but says nothing about the integration. Worth flagging because this PR drew three separate Codex review passes and none of them surfaced it — one lineage reviewing itself three times is one review, not three.

**To land it, in this order:** rename the flag in `openadapt-flow` (keeping `--break-it` as a deprecated alias there too), change its run dir to `run-rejected-write`, release it, raise the floor here to that version, then reopen. The branch `codex/quickstart-simulate-rejected-write` stays, so nothing is lost — reopen this PR rather than rewriting it.

Not closing for lack of merit. Closing because Codex is stopped, the Flow half does not exist, and an open PR that cannot merge is not a state worth holding during the drain.

Related: `openadapt-desktop#135` was grouped with this as one first-run cluster. It has no technical coupling to this PR — no reference to quickstart, the flag, or this number anywhere in its diff — so I landed it on its own merits rather than holding it behind this blocker.

